PR #32, opened 2021-08-23
Clamp highlight length to at least 1

### LINKED ISSUE #31: Printing errors with an empty highlight span panics.
```
thread 'single_line_highlight' panicked at 'attempt to subtract with overflow', .../miette/src/printer/graphical_printer.rs:388:29
```

I can reproduce this with the existing test suite by changing the 4 to 0 here: https://github.com/zkat/miette/blob/7e76e2dea4adf0e4a1349e049495c1f5a0bdab87/tests/printer.rs#L44

martica:
Fixes #31

### DIFF
--- a/src/printer/graphical_printer.rs
+++ b/src/printer/graphical_printer.rs
@@ -382,12 +382,13 @@ impl GraphicalReportPrinter {
         let mut highest = 0;
         let chars = &self.theme.characters;
         for hl in single_liners {
+            let hl_len = std::cmp::max(1, hl.len());
             let local_offset = hl.offset() - line.offset;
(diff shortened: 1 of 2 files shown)


zkat:
don't worry, that failure isn't your fault.

zkat (commented):
This looks good, but do you mind if I ask you to rebase this later? I'm in the middle of a big rewrite (shouldn't take me much longer!) and I think I can fix this doc error real quick once I'm done. Unless you want to fix it yourself?

zkat:
Yay thank you! I fixed stuff in main so you're all set :)

PR #39, opened 2021-08-30
Show snippet message for unnamed sources

### LINKED ISSUE #38: Snippet message not rendered when source is not named
This goes back to #23, when the "header" concept was introduced. The header is not rendered at all when the source returns None for `.name()`. (Back in v0.12 you needed SourceSpan::new_labeled.)

Workaround atm is to give every source a name. But given Source is implemented for unnamed `String`/`Arc<String>` miette should be able to render the header without a name.

Using `String`/ `Arc<String>`

```
Error: ────[code::here]────────────────────

    × std::error::Error display message

 1 │ one two thre
   ·     ─┬─
   ·      ╰── highlight message
```

Using NamedSource

```
Error: ────[code::here]────────────────────

    × std::error::Error display message

   ╭───[example.txt:1:1] snippet message:
 1 │ one two thre
   ·     ─┬─
   ·      ╰── highlight message


```

Now that I figured out the issue, it's like a few lines to edit across a few printers, so lemme give it a go.

While I'm thinking about this, I think we could implement Source on `&'static str` because these can be cloned just as freely as the others and are obviously `'static`. Not super useful except for testing.

cormacrelf:
Fixes #38

### DIFF
--- a/tests/printer.rs
+++ b/tests/printer.rs
@@ -510,5 +510,16 @@ fn unnamed_snippet_shows_message() {
     };
     let out = fmt_report(err.into());
     println!("{}", out);
-    assert!(out.contains("This is the part that broke"));
+    let expected = r#"
+────[oops::my::bad]────────────────────
+
+    × oops!
+
+   ╭───[1:1] This is the part that broke:
+ 1 │ source_text_here
+
+    ‽ try doing it better next time?
+"#
+    .trim_start();
+    assert_eq!(out, expected);
 }
--- a/tests/printer.rs
+++ b/tests/printer.rs
@@ -491,3 +491,24 @@ fn disable_url_links() -> Result<(), MietteError> {
     assert!(out.contains("oops::my::bad"));
     Ok(())
 }
+
+#[test]
+fn unnamed_snippet_shows_message() {
+    #[derive(Debug, Diagnostic, Error)]
+    #[error("oops!")]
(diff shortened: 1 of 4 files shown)


tests/printer.rs
```diff
@@ -491,3 +491,24 @@ fn disable_url_links() -> Result<(), MietteError> {
...
+    }
+    let src = "source_text_here".to_string();
+    let len = src.len();
+    let err = MyBad {
+        src,
+        ctx: (0, len).into(),
+    };
+    let out = fmt_report(err.into());
+    println!("{}", out);
+    assert!(out.contains("This is the part that broke"));
```
zkat, line 513: Could you please compare against the entire, rendered message. I made an exception for URLs because these tests (intentionally) don't test the escape codes themselves.

zkat (changes requested):

tests/printer.rs
cormacrelf: Yep

zkat:
haha cargo fmt to the rescue 😅

tests/printer.rs
cormacrelf: For another day, this might be a job for https://lib.rs/insta, to do what you've basically been doing manually in `tests/narrative.rs`

zkat: yes that would be great!

zkat:
yay thanks :D

PR #41, opened 2021-08-30
Add `#[diagnostic(forward(field_name), code(...))]`

cormacrelf:
This PR solves a problem I imagined. Not sure if it's real or not yet. I made a demo though: https://gist.github.com/cormacrelf/09ef573df0f5b41c9fa252f5866336ba

There are bugfix components of this that unquestionably need to be merged, but they were done on top of the other work so I'd have to redo it basically.

As described in #36:

> The forwarding branch, in addition to that, adds a forward(field_name) arg that must be used with a code(...) like any other arg. When you supply it, it makes any unspecified things forward their implementation to the specified field. It works with named and unnamed (forward(0)) structs and enum variants. The test suite shows what you can do with it -- the ForwardsTo struct is used for the snippets, all the other bits of info can be overridden.
>
> The question before we land this PR is a bikeshed, whether these two different argument names for arguably the same functionality are a good idea or not. Which is preferable of these?
>
> -    Only diagnostic(transparent)
> -    diagnostic(transparent), diagnostic(code(...), forward(field_name)) as described
> -    diagnostic(transparent), diagnostic(code(...), transparent(field_name))
> -    diagnostic(transparent), diagnostic(code(...), default(field_name). I kinda like this but I only just thought of it.


zkat:
quick comment: I went to some lengths to make sure that all the `gen` methods were infallible, so any issues would be found out as early as possible, during the parsing phase. I haven't read this PR in detail yet, but it would be nice if we could keep those infallible generators.

cormacrelf:
Ah yes, originally I peppered a few Results in there, but realised what you were going for later. I just forgot to remove the Result from the main gen() method. It did indeed turn out cleaner with the fallible parts in the parse phase.

cormacrelf:
One side benefit is that for use of transparent on structs with !=1 field, what used to be a rendered `_ => compile_error!("...")`  is now a properly spanned error, since it's now `DiagnosticDefArgs::Transparent(Forward)` constructed with `Forward::for_transparent_field()?`.

zkat:
hey how's it going with this? Just a heads up that I'm about to start making some potentially really big breaking changes to the way snippets are defined and I'm afraid to conflict with this stuff.

cormacrelf:
I did mark it draft, but there's nothing left to do.

cormacrelf:
Basically it's whether you want it or not!

cormacrelf:
Actually there are some conflicts now, thumbs up react if you want me to resolve em

zkat:
Please go ahead

cormacrelf:
The failing tests are the same ones that are failing on main due to a change in the number of `-------` after printing the code

zkat:
yay! Great job, and thanks :)

PR #42, opened 2021-08-30
Remove bound `T: Clone` from `Source` implementation for `Cow`.

olson-sean-k:
The bound `T: Clone` on [the `Source` implementation for `Cow<'_, T>`](https://github.com/zkat/miette/blob/84219f6c80c2c432fbeb4c40a591380285de8767/src/source_impls.rs#L87) is too restrictive. `Cow<'_, T>` [implements `Clone` using `T: ToOwned`](https://github.com/rust-lang/rust/blob/6cfa773583bb5123e630668f5bfe466716225546/library/alloc/src/borrow.rs#L199) and `T` need not implement `Clone` itself. This change relaxes the bounds so that the `Source` implementation includes `Cow<'_, str>`. Please take a look. Thanks!

### DIFF
--- a/src/source_impls.rs
+++ b/src/source_impls.rs
@@ -1,7 +1,11 @@
 /*!
 Default trait implementations for [Source].
 */
-use std::{borrow::{Cow, ToOwned}, fmt::Debug, sync::Arc};
+use std::{
+    borrow::{Cow, ToOwned},
+    fmt::Debug,
+    sync::Arc,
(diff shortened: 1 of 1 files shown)


zkat (approved):
Hey this is great! Thanks for this! I always found this `Cow: Clone` constraint obnoxious. I didn't realize it was just on `ToOwned`!

PR #52, opened 2021-09-12
Fix end of previous line wrongly being included in highlight

icewind1991:
when the highlight span starts at the very begining of a line, the highlight currently includes the line above

Before:

image

After:

image

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -592,7 +592,7 @@ impl Line {
 
     fn span_applies(&self, span: &FancySpan) -> bool {
         // Span starts in this line
-        (span.offset() >= self.offset && span.offset() <= self.offset +self.length)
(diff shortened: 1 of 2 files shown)


zkat:
oh sweet. Nice catch!

zkat:
ugh, looks like this does break some tests. These rules have been really tricky to get right :(

icewind1991:
Fixed the tests, the changes in highlight border box changed the frame width

zkat (approved):

zkat (changes requested):
The headers you changed all have to be 80 characters long, because they're meant to span the entire width of the terminal. Something funky's going on if they're 7 characters shorter now. Can you look into that?

zkat:
Nevermind that part is my fault. Can you remove the diffs on the headers and I'll just merge this? And can you add a test that covers this corner case?

icewind1991:
done

PR #82, opened 2021-10-03
Add replace tabs with spaces option

### LINKED ISSUE #73: Weird tab character rendering
I'm using `miette` in a parser that deals with tab `\t` indents and so far noticed the following:

* `miette` renders one `\t` as the number of spaces configured in the terminal (OK)
* `miette` *counts* one `\t`'s width as equal to one space and displays the arrow/label according to this width (Breaks because of the above, but should be possible to deal with by the developer)

So I decided to calculate number of tabs in the line, multiply it by tab width, and add that to the offset minus the number of tabs. Example: 

1. `\t\t` with tab width 4 = 8
2. Offset of some char that caused the error is 10
3. We add 10 + 8 and subtract 2 (number of tabs) and give that number as `SourceSpan` (16, 0)

In my repro this places the arrow in the correct position. But weird stuff happens if I have a really big offset (exactly 37 and more chars). It either just renders the line without rendering arrow/label or simply panics with this error:

```
thread 'main' panicked at 'failed printing to stderr: formatter error', library/std/src/io/stdio.rs:935:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Here's a reproduction:
https://github.com/jlkiri/miette-tab-error-repro

jlkiri:
Closes #73 

Makes it possible to replace tabs with spaces. This is needed because highlight line renderers are not aware of displayed tab width and appear in wrong places if line contains tabs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,7 @@ miette::set_hook(Box::new(|_| {
         .terminal_links(true)
         .unicode(false)
         .context_lines(3)
+        .with_tab_width(4)
         .build())
 }))
 
--- a/README.md
+++ b/README.md
@@ -389,7 +389,7 @@ miette::set_hook(Box::new(|_| {
         .terminal_links(true)
         .unicode(false)
         .context_lines(3)
-        .with_tab_width(4)
+        .tab_width(4)
         .build())
 }))
 
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -565,6 +565,7 @@ impl GraphicalReportHandler {
             let hl_len = std::cmp::max(1, hl.len());
 
             let local_offset = if let Some(w) = self.tab_width {
+                // Only count tabs that affect the position of the highlighted line and ignore tabs past the span.
                 let tab_count = &line.text[..hl.offset() - line.offset].matches('\t').count();
                 let tabs_as_spaces = tab_count * w - tab_count;
                 hl.offset() - line.offset + tabs_as_spaces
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -389,6 +389,7 @@ miette::set_hook(Box::new(|_| {
         .terminal_links(true)
         .unicode(false)
         .context_lines(3)
+        .with_tab_width(4)
```
zkat, line 392: this should be just `.tab_width`

src/handlers/graphical.rs
```diff
@@ -48,9 +50,16 @@ impl GraphicalReportHandler {
             theme,
             footer: None,
             context_lines: 1,
+            tab_width: None,
         }
     }
 
+    /// Replace tabs with spaces.
+    pub fn with_tab_width(mut self, width: usize) -> Self {
```
zkat, line 58: You need to expose `.tab_width()` through `MietteHandler`, too, since that's the recommended way of customizing your handler in general.

tests/test_location.rs
```diff
@@ -30,7 +30,6 @@ impl miette::ReportHandler for LocationHandler {
     }
 
     fn track_caller(&mut self, location: &'static Location<'static>) {
-        dbg!(location);
```
zkat, line 33: oops! :P

zkat (changes requested):

src/handlers/graphical.rs
jlkiri: Done cb073dd3e773e5ac39db7d3ac811fcbf6c53e82d

README.md
jlkiri: Done cb073dd3e773e5ac39db7d3ac811fcbf6c53e82d

src/handlers/graphical.rs
```diff
@@ -25,6 +25,7 @@ pub struct GraphicalReportHandler {
     pub(crate) theme: GraphicalTheme,
     pub(crate) footer: Option<String>,
     pub(crate) context_lines: usize,
+    pub(crate) tab_width: Option<usize>,
```
dswij, line 28: This might sound dumb, but I wonder if a user should be allowed to use `tab_width = 0` for no space or it'd rather be better if `tab_width = 0` means using tabs 🤔

zkat (approved):

Issue #29, opened 2021-08-21
Machine-parseable error reporter


zkat:
Might be nice to have a reporter that's able to output errors as JSON.

Ref: https://github.com/zkat/miette/issues/28#issuecomment-903153221.

dvermd:
Hi, I've started working on a JSONReporter. 

Were you thinking of pulling serde_json dependency or going to a hand crafted serialisation ?

zkat:
I think doing hand-serisalization is sufficient in this case, and it avoids an extra dependency. What do you think?

dvermd:
I've got most of the serialization done. Only `SourceCode` is left. 
Do you think `SourceCode` made it to a machine-writable format ?

In case, yes:
Both `NarratableReportHandler` and `GraphicalReportHandler` use the same `render_snippets` function that calls `render_context` function which each reporter has its own.

A utility function can be extracted from `render_snippets` and be used in the 3 Reporters  but I wonder where would be the best place to put this function.

I don't think the creation of the JSONReporter has its place in the MietteHandlerOpts builder. It can be instantiated directly or with its own builder to provide `context_lines`. What's your view on that ?

Issue #88, opened 2021-10-15
Tests in README failing without fancy feature


LordMZTE:
When running `cargo test` without `--features fancy`, one of the tests in `README.md` fails:
```
failures:

---- src/lib.rs - (line 386) stdout ----
error[E0433]: failed to resolve: could not find `MietteHandlerOpts` in `miette`
 --> src/lib.rs:388:22
  |
5 |     Box::new(miette::MietteHandlerOpts::new()
  |                      ^^^^^^^^^^^^^^^^^ could not find `MietteHandlerOpts` in `miette`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0433`.
Couldn't compile the test.

failures:
    src/lib.rs - (line 386)

test result: FAILED. 26 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.86s

error: test failed, to rerun pass '--doc'
```

The error is a little confusing due to the `include_str!` used in lib.rs, but it's caused by line 388 in `README.md`.
Looks like the code is referencing a struct that exists only with the fancy feature. Maybe we should add a `cfg` attribute to the example function?
We should also consider adding a step to the CI that runs tests without the fancy feature.

I can also open a PR to fix this :D

PR #90, opened 2021-10-18
feat(handlers): Add JSON handler

dvermd:
Closes: #29


dvermd:
As of #88, tests are to be run with `cargo test --features=json,fancy --all`

zkat (approved):

PR #100, opened 2022-01-06
docs: fix mentions of eyre and DiagnosticError

tailhook:
Note: `DiagnosticError` is a private type, `Report` returned from
`into_diagnostic()` instead.


zkat (approved):
aaaahh thanks. Guess I missed some spots there :)

PR #102, opened 2022-01-07
fix(chain): correct `Chain` structure exported

tailhook:
This fixes all current compiler and clippy warnings.

It looks like wrong `Chain` type was exported. This is technically could be seemed as a breaking change, but in reality should not break anyone since that type was not used and could not be created I think.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,6 @@
 
 pub use miette_derive::*;
 
-pub use chain::*;
 pub use error::*;
 pub use eyreish::*;
 #[cfg(feature = "fancy")]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 pub use miette_derive::*;
 
+pub use chain::*;
```
zkat, line 7: Is this necessary?

tailhook: Yes, I think exporting named opaque type in [chain method](https://docs.rs/miette/latest/miette/struct.Error.html#method.chain) is better than exporting private type.

Alternative is to change `chain` to return `impl Something`.

Issue #97, opened 2021-12-16
Label location and width do not account for character length


Kylebrown9:
Label locations do not correctly account for the actual width of characters.

The specific case I tested was for `☃` which is treated as being 3-wide because it is 3 bytes in utf-8.

Here is a snippet of output for the issue.
```
LexerError { src: NamedSource { name: "bad.wrt", source: "<redacted>", span: SourceSpan { offset: SourceOffset(4), length: SourceOffset(3) } }

  × The input did not match a token rule
   ╭─[bad.wrt:1:1]
 1 │ abc ☃ abc
   ·     ─┬─
   ·      ╰── This text was not recognized
   ╰────
```

tasogare3710:
This is a common issue and you need to comply [UAX #29](http://www.unicode.org/reports/tr29/).

Aloso:
This should be possible to fix with the [unicode-width](https://docs.rs/unicode-width/latest/unicode_width/) crate.

tasogare3710:
This issue requires an accurate count of the user-perceived character(s)(a.k.a Grapheme Cluster), and UAX11 has nothing to do with this issue.

See also http://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries

Aloso:
@tasogare3710 It does not require an exact count of the user-perceived characters, it requires an exact **width** of the user-perceived characters in a terminal. Since some characters, such as emojis and East Asian characters, are usually 2 columns wide instead of 1, label locations will still be incorrect if you just count grapheme clusters.

PR #107, opened 2022-01-17
fix(handlers): Fix label position

### LINKED ISSUE #106: Are labels in narratable report handler right?
Here is the snippet from the test:
```rust
    let src = "source\n  👼🏼text\n    here".to_string();
    let err = MyBad {
        src: NamedSource::new("bad_file.rs", src),
        highlight: (9, 6).into(),
    };
    let out = fmt_report(err.into());
    println!("Error: {}", out);
    let expected = r#"oops!
    Diagnostic severity: error
Begin snippet for bad_file.rs starting at line 1, column 1
snippet line 1: source
    label starting at line 1, column 1: this bit here
snippet line 2:   👼🏼text
snippet line 3:     here
diagnostic help: try doing it better next time?
diagnostic code: oops::my::bad
"#
```
The highlight (9, 6) should be on line 2 column 3 I think. No? Actually I've noticed this on my own example which is very similar, the span is on the middle of the 18th line, where snippet starts from 17th. But it prints "label starting at line 17, column 1" (so looks like sticks always to the first column of the first displayed line).

Am I misunderstanding narratable handler?

tailhook:
Fixes #106

While fixing the bug I've also extended the handler to print not just
where label starts but also where label ends.

Also implements #97, but for narratabble handler only.

This adds dependency on `unicode-width`, but it was already in the
dependency chain (`textwidth` depends on it).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ exclude = ["images/", "tests/", "miette-derive/"]
 thiserror = "1.0.26"
 miette-derive = { path = "miette-derive", version = "=3.3.1-alpha.0"}
 once_cell = "1.8.0"
+unicode-width = "0.1.9"
 
 owo-colors = { version = "3.0.0", optional = true }
 atty = { version = "0.2.14", optional = true }
@@ -45,7 +46,7 @@ fancy = [
     "supports-hyperlinks",
     "supports-color",
     "supports-unicode",
-    "backtrace"
+    "backtrace",
 ]
 
 [workspace]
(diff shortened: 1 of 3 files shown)


tailhook:
Had to make the `unicode-width` dependency to non-optional, since `NarratableErrorHandler` is always supported. Let me know if you'd like to make the handler be supported in `fancy` mode only (although, that would make it a breaking change).

zkat (approved):
oh hey this is great! Thank you!

I think adding unicode-width as non-optional is ok--it's pretty small and non-invasive.

PR #113, opened 2022-02-12
Add field variant of help

### LINKED ISSUE #112: Dynamic help text
I have an error which can have different help text depending on the underlying I/O error and the OS. At the moment I'm abusing `#[source_code]` to render help text, which isn't great.

I could make different error variants for each I/O error and use `cfg_attr` to vary on OS:

```rust
enum Ohno {
	#[cfg_attr(target_os = "linux", help("Try increasing inotify instance limits or your nofile user limits")]
	TooManyHandles,

	#[cfg_attr(target_os = "linux", help("Try increasing inotify watch limits")]
	TooManyWatches,

	// ...
}
```

but it gets a bit out of hand.

Would it be in scope to support e.g.

```rust
struct Ohno {
	#[source] inner: std::io::Error,
	#[help] help: String, // or even `Option<String>`
}
```

passcod:
Fixes #112.  Looks like:

```rust
#[derive(Debug, Diagnostic, Error)]
#[error("welp")]
struct FooStruct {
    #[help]
    help: &'static str,
}

#[derive(Debug, Diagnostic, Error)]
#[error("welp")]
enum FooEnum {
    X(#[help] String),

    Y {
        #[help]
        help: String,
    },
}
```

I'd like some help!

- [x] General "is this the right direction"
- [ ] Forwarding-with-overriden-help tests are failing, and I'm not sure how to approach it. I'll keep looking, but would appreciate any pointers
- [x] When both derive-arg and field help is provided, the procmacro panics because this is detected at codegen time. Is this fine? Is there a better way to handle it?

### DIFF
--- a/miette-derive/src/diagnostic.rs
+++ b/miette-derive/src/diagnostic.rs
@@ -4,6 +4,7 @@ use syn::{punctuated::Punctuated, DeriveInput, Token};
 
 use crate::code::Code;
 use crate::diagnostic_arg::DiagnosticArg;
+use crate::field_help::FieldHelp;
 use crate::forward::{Forward, WhichFn};
 use crate::help::Help;
 use crate::label::Labels;
@@ -63,6 +64,7 @@ pub struct DiagnosticConcreteArgs {
     pub help: Option<Help>,
     pub labels: Option<Labels>,
     pub source_code: Option<SourceCode>,
+    pub field_help: Option<FieldHelp>,
     pub url: Option<Url>,
     pub forward: Option<Forward>,
     pub related: Option<Related>,
@@ -105,6 +107,7 @@ impl DiagnosticConcreteArgs {
         }
         let labels = Labels::from_fields(fields)?;
         let source_code = SourceCode::from_fields(fields)?;
+        let field_help = FieldHelp::from_fields(fields)?;
         let related = Related::from_fields(fields)?;
         let concrete = DiagnosticConcreteArgs {
             code,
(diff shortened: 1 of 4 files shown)


zkat:
1. Yes this looks good to me.
2. Which tests? I don't actually remember how this part worked haha
3. Yes that is fine.

passcod:
```rust
#[test]
fn test_forward_enum_named() {
    #[derive(Debug, Diagnostic, Error)]
    enum Enum {
        #[error("help: {help_text}")]
        #[diagnostic(code(foo::bar::overridden), help("{help_text}"), forward(span))]
        Variant {
            span: ForwardsTo,
            help_text: &'static str,
        },
    }
    // Also check the From impl here
    let variant: Enum = Enum::Variant {
        span: ForwardsTo::new(),
        help_text: "overridden help please",
    };
    assert_eq!(variant.code().unwrap().to_string(), "foo::bar::overridden");
    assert_eq!(
        variant.help().unwrap().to_string(),
        "overridden help please"
    );
}
```

fails with

```
thread 'test_forward_enum_unnamed' panicked at 'assertion failed: `(left == right)`
  left: `"help"`,                                                                  
 right: `"overridden help please"`', tests/derive.rs:528:5
```

(forward is being too eager with my changes?)

zkat:
@cormacrelf mind if I drag you in here to help with the Forward stuff?

passcod:
Actually, nevermind this. While looking at the code I've realised what I want is already possible!

PR #115, opened 2022-02-15
Make derive macro `diagnostic` attribute more flexible.

Kylebrown9:
When deriving `Diagnostic` for an enum, if the attribute `diagnostic` is not provided on the variant, the macro will attempt to use one defined on the enum itself.

This should allow users to place the diagnostic attribute in either location.
The more specific attribute will still win, which could be useful in the case that a user defines a broad one for the whole enum and a specific one for variants that need it.

### DIFF
--- a/miette-derive/src/diagnostic.rs
+++ b/miette-derive/src/diagnostic.rs
@@ -155,9 +155,10 @@ impl DiagnosticDefArgs {
 
 impl Diagnostic {
     pub fn from_derive_input(input: DeriveInput) -> Result<Self, syn::Error> {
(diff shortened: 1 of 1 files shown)


Kylebrown9:
I've expanded on the original scope based on discussions with @zkat.
In addition to allowing you to specify the `#[diagnostic(...)]` attribute on either the variant or the item you can
* Not provide it at all
* Provide it on both levels and have it combine the arguments from them
* Provide it multiple times at either or both levels with it combining well

zkat (approved):
this is so great! LGTM and thank you for the patch!

Issue #99, opened 2022-01-05
Can renderer of related errors (optionally) preserve source code?


tailhook:
The scenario is like:
```
struct ParentError {
    #[source_code]
    source: String,
    #[related]
    other: Vec<ChildError>,
}
```
I can use `Arc` to copy source code to every child error, but this is a bit inconvenient. Parsers usually only propagate byte range as a span so I have to either wrap every error into another wrapper having source code, or propagate some custom span that includes `Arc<SourceCode>`.

Can handler propagate that source code while rendering or are there any big downsides of that?

zkat:
This can be done pretty easily [by using an error wrapper that "forwards" the `source_code` implementation](https://github.com/nushell/engine-q/blob/8cf4402e6cbfbe850ee47521bedb65c890baba89/crates/nu-cli/src/errors.rs)

Would that work for you? It seemed to work pretty well for engine-q.

tailhook:
This means I should make a manual wrapper with a `related` method kinda like this:
```rust
fn related<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a dyn Diagnostic> + 'a>> {
  Some(self.other.iter().map(|e| ErrorWrapper { source: &self.source, error: e }))
}
```
But it's impossible to do as iterator should return a reference and I can get a reference for the temporary object like above. So it will need to end up something like this:
```rust
struct ParentError {
    #[source_code]
    source: Arc<String>,
    #[related]
    other: Vec<ErrorWrapper<ChildError>>,
}
```
Still having an Arc to clone the source code into every error.

PR #117, opened 2022-02-18
More ergonomic source code handling

### LINKED ISSUE #64: `.source_code()` and `.with_source_code()` methods for `Report`
This will make it possible to delay providing `source_code` until the app is ready to provide it, which is a common pattern in compilers.

tailhook:
Fixes #64 and #99

@zkat, I'm not sure what you mean by `source_code()` and `with_source_code()` code in #64. Since the report object is already an error it doesn't make much sense to put a lambda there. And since `source_code()` might be confused with `Diagnostic::source_code` I think `with_` works better.

The internal `WithSourceCode` wrapper is designed in a way that would allow adding `with_source_code` on error or even `WrapErr`. But I'm not sure these are needed, so did not include in the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ and such might not want.
   - [... diagnostic code URLs](#-diagnostic-code-urls)
   - [... snippets](#-snippets)
   - [... multiple related errors](#-multiple-related-errors)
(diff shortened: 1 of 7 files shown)


tailhook:
Added a couple of examples into the README

PR #121, opened 2022-02-22
feat(derive): allow `Report` in `related`

### LINKED ISSUE #119: Use `Report` in `related`
Currently you can't put `Vec<Report>` into the `related` field, only concrete error. This is quite inconvenient in certain cases.

One option would be to make some way to agument `related`:
```rust
struct MultiError {
  #[related(Report)]
  related: Vec<Report>,
}
```
Or maybe more generic:
```rust
struct MultiError {
  #[related(AsRef)] // or #[related(as_ref)]
  related: Vec<Report>,
}
```
Or even more generic:
```rust
struct MultiError {
  #[related(|x| x.as_ref())]
  related: Vec<Report>,
}
```

It should also be possible to use [dtolnay's hack](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md). But I'm not sure how much complexity that adds to the derive.

What do you think?

tailhook:
Fixes #119

I'm not sure if using `borrow()` should be thought of as a great way to handle the issue or it's a hack.

The alternative would be to make some private trait that we have to expose using `docs(hidden)` for derive to use. But this may also make error reporting confusing (i.e. failing with trait name that can't be found anywhere in the docs).

I think `Borrow` is safe to implement for the `Report`. I had a tiny concern that it make prohibit using error types with (non-static) lifetimes in `related`. But I think it's *not the case*. Although, I could not test that as code-gen for nonstatic lifetimes in `related` is untested and broken for now.

### DIFF
--- a/miette-derive/src/related.rs
+++ b/miette-derive/src/related.rs
@@ -69,7 +69,10 @@ impl Related {
         let rel = &self.0;
         Some(quote! {
             fn related<'a>(&'a self) -> std::option::Option<std::boxed::Box<dyn std::iter::Iterator<Item = &'a dyn miette::Diagnostic> + 'a>> {
-                std::option::Option::Some(std::boxed::Box::new(self.#rel.iter().map(|x| -> &(dyn miette::Diagnostic) { &*x })))
+                use ::core::borrow::Borrow;
+                std::option::Option::Some(std::boxed::Box::new(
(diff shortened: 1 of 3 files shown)


zkat (approved):
Let's give this a whirl and see how it goes! :) I think this'll make a lot of things easier in practice, even if we run into some pain points later.

zkat:
Tagged and released as 4.2.0

PR #126, opened 2022-03-05
Relax hook to work on non-'static diagnostics

CAD97:
This is a fundamental tradeoff. I think, for diagnostics, while `Report` _should_ be `'static` as an error reporting type, individual `Diagnostic`s have good reason to be non-`'static`, and the library should try to support that. For example, I have a `ParseError<'src>` type which contains `&'src str` to the parsed source and implements `Diagnostic`.

The loss here is that the error hook can no longer `Error::downcast` the diagnostic. For eyre, it makes some sense to customize the error display style based on the concrete type of the error; perhaps there's an error type that you want a more friendly error formatting handler for, and you want the default for other errors. For miette, though, downcasting and choosing a different report handler seems ill-advised at best. (Plus, you can always just not use the global hook and feed it through a report handler directly.)

...And now that `downcast` isn't available, you could even just have `set_hook(&'static dyn ReportHandler)`, and skip the creation of a new report handler for each error. (See the discussion #128 for my more in detail opinions here.)

### DIFF
--- a/src/eyreish/mod.rs
+++ b/src/eyreish/mod.rs
@@ -54,8 +54,7 @@ pub struct Report {
 }
 
 ///
-pub type ErrorHook =
-    Box<dyn Fn(&(dyn Diagnostic + 'static)) -> Box<dyn ReportHandler> + Sync + Send + 'static>;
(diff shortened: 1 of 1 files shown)


CAD97:
(Closing for #128 discussion)

PR #133, opened 2022-03-18
handler_opts.color(false) should disable color

CAD97:


zkat (approved):
Ooh nice catch. Thank you!

zkat:
This has been released as part of miette@4.3.0

PR #137, opened 2022-03-18
Allow GraphicalReportHandler to disable url display

### LINKED ISSUE #136: Add option to fully disable url output (for tests)
As an [example snapshot test](https://github.com/CAD97/tracing-filter/blob/a68a6c5a3ea7a91389f94ba270df0d699eacd8a8/tests/snapshots/lib__snapshots__report%40invalid-filter.env_filter.snap):

```snap
---
source: tests/snapshots.rs
assertion_line: 30
expression: "crate::target=warn/[0-9]/"
---
tracing_filter::simple::Error::MultipleSlash (https://docs.rs/tracing-filter/0.0.0-dev/tracing-filter/enum.Error.html#variant.MultipleSlash)

  × logging spec has too many `/`s
   ╭────
 1 │ crate::target=warn/[0-9]/
   ·                    ───┬──
   ·                       │ ╰── this `/` is not allowed ...
   ·                       ╰── ... in this regex filter
   ╰────
  help: regex filters may not contain `/`

```

Because I've used `#[diagnostic(url(docsrs))]`, the docsrs URL is embedded into the diagnostic, and the output is version-dependent. This means that bumping the version number will also require re-blessing all of the snapshot tests with the new version number.

I'm already directly using `GraphicalReportHandler` to get (hopefully) environment-independent formatting with no ANSI codes

<details>

```rust
impl fmt::Display for DisplayDiagnostic<'_> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        miette::GraphicalReportHandler::new_themed(miette::GraphicalTheme::unicode_nocolor())
            .with_links(false)
            .render_report(f, self.0)
            .or_else(|_| write!(f, "failed to pretty-print {:#?}", self.0))
    }
}
```
</details>

but it'd be nice to be able to also completely remove the URL for test purposes.

This would add a new flag to `GraphicalReportHandler`, `with_urls(urls: bool)`, which would fully disable reporting of `Diagnostic::url`.

CAD97:
Closes #136

I've done my best to match existing code style.

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -22,7 +22,7 @@ printer.
 */
 #[derive(Debug, Clone)]
 pub struct GraphicalReportHandler {
-    pub(crate) linkify_code: bool,
+    pub(crate) links: LinkStyle,
     pub(crate) termwidth: usize,
     pub(crate) theme: GraphicalTheme,
     pub(crate) footer: Option<String>,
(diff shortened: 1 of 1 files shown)


zkat (approved):

zkat:
This has been released as part of miette@4.3.0

PR #150, opened 2022-04-14
fix: set correct field in MietteHandlerOpts::ansi_colors

sunshowers:
Thanks for miette! We've been having a great time using it in
cargo-nextest to print out nicely-formatted errors.

I noticed a bug while testing out miette -- the `ansi_colors` method
does not set ANSI colors properly.

By the way, I do not believe miette should use RGB colors at all --
default colors in applications must be restricted to 12 ANSI colors. See
https://rust-cli-recommendations.sunshowers.io/colors.html#color-palettes
for why. I'm happy to disable RGB colors in all the applications I
maintain, but doing so globally is something for upstream to consider.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -93,7 +93,7 @@ impl MietteHandlerOpts {
 
     /// If true, ANSI colors will be used during graphical rendering.
     pub fn ansi_colors(mut self, color: bool) -> Self {
-        self.rgb_colors = Some(color);
+        self.ansi_colors = Some(color);
         self
     }
     /// If true, graphical rendering will be used regardless of terminal


zkat (approved):
wow, copy/paste is incredibly rude. Absolutely terrible.

Thanks for this patch!

zkat:
also re: ansi colors

That's a pity, re: ansi colors! I spent a good amount of time making sure the default colors were both attractive _and_ colorblindness friendly, both of which I think the default ANSI color set kinda... fails to do in most cases.

That said, perhaps you're right and it shouldn't be the default. I'd be happy to take a patch that makes ANSI the default, but the specifics might warrant some discussion. Happy to see a concrete proposal, though!

sunshowers:
Thanks for considering the change to ANSI colors!

> That's a pity, re: ansi colors! I spent a good amount of time making sure the default colors were both attractive and colorblindness friendly, both of which I think the default ANSI color set kinda... fails to do in most cases.

I think that is definitely true for the default color scheme in most terminals, but the benefit of them is that users can customize them arbitrarily. I would imagine that most colorblind users would have set up an appropriate theme already (though I don't have data to support this!).

I think (as in the color palette link from my recommendations above) that the benefit of being able to customize colors in a global fashion is precisely the benefit that ANSI colors offer and that's harder to access with RGB colors.

Issue #160, opened 2022-04-23
IntoDiagnostic occludes possible 'Diagnostic' implementations


TheNeikos:
The following test fails:

```rust
use miette::{Diagnostic, IntoDiagnostic};
use thiserror::Error;

#[derive(Debug, Error, Diagnostic)]
#[error("Just another diagnostic")]
struct AnotherDiagnostic;

#[derive(Debug, Error, Diagnostic)]
#[error("Diagnostic with a related error")]
struct DiagnosticWithRelated {
    #[diagnostic(related)]
    rel: Vec<AnotherDiagnostic>,
}

#[test]
fn test_diagnostic_not_losing_related() {
    let dwr: Result<(), _> = Err(DiagnosticWithRelated {
        rel: vec![AnotherDiagnostic],
    });

    let dwr = dwr.into_diagnostic();

    assert!(
        dwr.err()
            .expect("Contain an error")
            .related()
            .expect("To have related if wrapping another diagnostic")
            .count()
            > 0
    );
}
```

Which is because `IntoDiagnostic` wraps a `std::error::Error` and then puts _that_ into a `miette::Report` which then has no idea whether the wrapped type might implement `Diagnostic`. 

This can cause confusion as one might expect that any `T: Diagnostic` when called `into_diagnostic` on would keep its original implementation (or atleast be forwarded to). 

I understand that this currently cannot be resolved without specialization due to overlapping impl with the `Error` implementation. (Although I am not 100% sure this can be solved with it either). 

Either way, I think that there are steps one could take to alleviate this problem somewhat:

- Add a warning that you should only call it on types that _only_ implement `Error` and not `Diagnostic`
- Deprecate the current name (or even doc(hide) it) and call it something else that makes it clear it shouldn't be used on `Diagnostic`s (although I don't think this is particularly effective)

PR #161, opened 2022-04-23
Expand documentation on IntoDiagnostic

TheNeikos:
This patch expands on the effects of using IntoDiagnostic and warns
users of potential loss of information when using it on `Diagnostic`
types.

This patch is related to #160.


zkat (approved):
This is as good a plan as any as I can think of right now. If you get any other ideas on how to address them, feel free to send them my way, for sure!

Issue #164, opened 2022-04-27
Relationship between Diagnostic and StdError


TheNeikos:
I would like to understand where miette currently tries to position itself in the 'Rust error story'.  To do this, I'm opening this issue to see what I might have missed and/or not understood correctly.

-------

Miette exposes a `Diagnostic` trait that extends on top of the `Error` trait with additional information so as to report better looking and more informative errors.

To do so, it currently imposes the following relationship `Diagnostic: StdError` (`StdError` is a supertrait of `Diagnostic`). This means that any `Diagnostic` is also an `StdError`.
`StdError` has multiple advantages that speak for it:
- It requires a 'Display' implementation, forcing developers to give a textual representation of the error that is fit for 'end-user consumption'.
- It has a `.source/cause` method that gives access to the error that 'cause'd it

`Diagnostic` builds on top of this with the ability to specify spans with labels, severity, etc... . 
This gives it several nice things that seperate it from `StdError`:
- Users can be more precisely informed with visual language on _where_ errors occured
- Colors (through severity) give them a color indicator on how 'bad' or useful the information is
- Through the `.source` method of `StdError` on can get access to the underlying errors and print those too, potentially going as far as to the root error that caused it.

This means that for a given `Diagnostic` and finally `Report` one can have the following relationship:

```

             ┌────────┐
     ┌───────┤ Report ├────┐
     │       └────────┘    │
     │                     │
     │                     │
┌────▼─────┐          ┌────▼─────┐
│Diagnostic│          │Diagnostic│
└────┬─────┘          └─────┬────┘
     │                      │
     │                      │
     │                      │
 ┌───▼────┐            ┌────▼───┐
 │StdError│            │StdError│
 └────────┘            └────┬───┘
                            │
                            │
                            │
                       ┌────▼───┐
                       │StdError│
                       └────────┘

```

Important to note is that while the Top-level Diagnostic/Report has multiple 'Diagnostics' below it, they are currently only 'related'. (For example multiple logically distinct errors in a configuration file).

It is also not possible to 'chain' Diagnostics, as there is no `.source` equivalent for `Diagnostic`. This means that either a single diagnostic is created at the root (potentially wrapping a StdError) and then transported all the way to the base or that extra information is added later on. (Through for example the `with_source_code` transforms). An error, once made into a `Diagnostic` cannot be re-integrated into the chain of errors except as a logically distinct 'related' issue.

-------

[...]

PR #165, opened 2022-04-28
Add caused_by to Diagnostic

matthiasbeyer:
Hi!

Currently, miette does not support error causes which contain a `Diagnostic`. 

This draft tries to change that. It is related to #164 , as I worked with @TheNeikos on implementing this.

Tell me what you think! :heart:

---
Please note that I will rebase this PR until we all agree on a way forward!

### DIFF
--- a/miette-derive/src/diagnostic_source.rs
+++ b/miette-derive/src/diagnostic_source.rs
@@ -1,0 +1,78 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::spanned::Spanned;
+
+use crate::forward::WhichFn;
+use crate::{
+    diagnostic::{DiagnosticConcreteArgs, DiagnosticDef},
+    utils::{display_pat_members, gen_all_variants_with},
+};
+
+pub struct DiagnosticSource(syn::Member);
+
+impl DiagnosticSource {
+    pub(crate) fn from_fields(fields: &syn::Fields) -> syn::Result<Option<Self>> {
+        match fields {
+            syn::Fields::Named(named) => Self::from_fields_vec(named.named.iter().collect()),
+            syn::Fields::Unnamed(unnamed) => {
+                Self::from_fields_vec(unnamed.unnamed.iter().collect())
+            }
+            syn::Fields::Unit => Ok(None),
+        }
+    }
+
+    fn from_fields_vec(fields: Vec<&syn::Field>) -> syn::Result<Option<Self>> {
+        for (i, field) in fields.iter().enumerate() {
+            for attr in &field.attrs {
+                if attr.path.is_ident("diagnostic_source") {
+                    let diagnostic_source = if let Some(ident) = field.ident.clone() {
+                        syn::Member::Named(ident)
+                    } else {
+                        syn::Member::Unnamed(syn::Index {
+                            index: i as u32,
+                            span: field.span(),
+                        })
+                    };
+                    return Ok(Some(DiagnosticSource(diagnostic_source)));
+                }
+            }
+        }
+        Ok(None)
+    }
+
+    pub(crate) fn gen_enum(variants: &[DiagnosticDef]) -> Option<TokenStream> {
+        gen_all_variants_with(
+            variants,
+            WhichFn::DiagnosticSource,
+            |ident,
+             fields,
+             DiagnosticConcreteArgs {
+                 diagnostic_source, ..
+             }| {
+                let (display_pat, _display_members) = display_pat_members(fields);
+                diagnostic_source.as_ref().map(|diagnostic_source| {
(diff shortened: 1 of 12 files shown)


src/handlers/graphical.rs
```diff
@@ -199,8 +200,8 @@ impl GraphicalReportHandler {
 
         writeln!(f, "{}", textwrap::fill(&diagnostic.to_string(), opts))?;
 
-        if let Some(cause) = diagnostic.source() {
-            let mut cause_iter = Chain::new(cause).peekable();
+        if let Some(cause) = diagnostic.caused_by() {
+            let mut cause_iter = DiagnosticChain::new(cause).peekable();
```
zkat, line 204: I want `#[source]` to continue working for folks who are already using that, or who don't want to opt into `#[caused_by]`. Thoughts? It seems pretty doable.

TheNeikos: IIUC, `DiagnosticChain` does do this: https://github.com/zkat/miette/pull/165/files/67b27eb94e80e27e4a4addc1a0a1bf03c4ed785f#diff-dd270622a98a63e007d36dd28857130489803df52855385b3150ef332ce4d344R61

It tries to stay on the `Diagnostic` path as long as it can

zkat: No, that assumes that at the very least the toplevel "source" is a `Diagnostic`, only following the `Error::source` trail _after_ the toplevel one has been resolved. I'm talking about the case where `#[caused_by]` is never defined at all, and `thiserror`'s `#[source]` is used instead (or a manual impl of `Error::source`).

TheNeikos: You're right! Totally missed that...

matthiasbeyer: I pushed a fix which should address this.

matthiasbeyer:
For me, tests are green now :tada:

matthiasbeyer:
Definitively. Tell me if there's more that you want to have changed. I'll make sure to squash the history appropriately before it gets merged, if it gets merged.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "miette"
-version = "4.6.1-alpha.0"
+version = "4.4.8"
```
TheNeikos, line 3: I think this should not be in the commit

matthiasbeyer: Whoopsi.

miette-derive/src/diagnostic_source.rs
```diff
@@ -63,14 +63,5 @@ impl CausedBy {
...
     }
-
-    pub(crate) fn gen_struct(&self) -> Option<TokenStream> {
-        let rel = &self.0;
-        Some(quote! {
-            fn diagnostic_source<'a>(&'a self) -> std::option::Option<&'a dyn miette::Diagnostic> {
-                std::option::Option::Some(self.#rel.as_ref())
-            }
-        })
-    }
```
zkat: uhhh, this shouldn't be unused.

The `gen_struct` typically generates definitions for when the macro is used on a struct-style error, for example:

```rust
#[derive(Diagnostic, Error)]
#[error("meh")]
struct MyError {
  #[diagnostic_source]
  src: SomeError
}
```

matthiasbeyer:
Okay, I guess now this looks like it is ready ... @zkat if you agree, I can squash and make the history here pretty before it goes in!

zkat:
you'll want all the tests passing first :)

tests/test_diagnostic_source_macro.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Debug, miette::Diagnostic, thiserror::Error)]
+#[error("TestError")]
+struct TestError {
+    #[diagnostic_source]
+    asdf_inner_foo: AnErr,
+}
+
+#[test]
+fn test_diagnostic_source() {
+    let error = TestError { asdf_inner_foo: AnErr };
```
zkat, line 16: re: test failures: this needs formatting. Try `cargo fmt --all`.

src/handlers/graphical.rs
```diff
@@ -3,6 +3,7 @@ use std::fmt::{self, Write};
 use owo_colors::{OwoColorize, Style};
 
 use crate::chain::Chain;
```
zkat, line 5: re: test failures: this is unused now.

zkat (approved):
everything looks good. I'll merge once tests are passing again :)

Thanks for the contribution, and for hanging in there during the back and forth!

matthiasbeyer:
Rebased and squashed everything no be logically grouped into commits. This includes #168 for now, but I am happy to rebase if #168 gets merged.

zkat:
yay! Thanks again! :) Looks like we're all set~

PR #168, opened 2022-05-05
Fix: Formatting

matthiasbeyer:
Found that `main` was not formatted. Fixed with this commit.


zkat (approved):
Weird. I wonder how this managed to land at all. I try not to merge stuff without it passing CI. Maybe there was a version update or something.

PR #170, opened 2022-05-06
Render inner diagnostics

TheNeikos:
Previously `#[diagnostic_source]` did not correctly render the inner diagnostic. This PR fixes that.

There's a change here that can be considered controversial:

- Only print newlines when there is an actual header (in ad9d0378ddd6dd4a6b9e650dd6aa62116da65910)

I am fine with removing it, but then the nested diagnostic source will be offset by one line and appear floating when it does not print a header.

PS: This PR also includes things from #169. I will rebase once that is merged (hopefully?)

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -261,6 +261,7 @@ impl GraphicalReportHandler {
                         // Don't print footer for inner errors
                         let mut inner_renderer = self.clone();
                         inner_renderer.footer = None;
+                        inner_renderer.with_cause_chain = false;
                         inner_renderer.render_report(&mut inner, diag)?;
 
                         writeln!(f, "{}", textwrap::fill(&inner, opts))?;
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -151,7 +151,6 @@ impl GraphicalReportHandler {
         diagnostic: &(dyn Diagnostic),
     ) -> fmt::Result {
         self.render_header(f, diagnostic)?;
-        writeln!(f)?;
         self.render_causes(f, diagnostic)?;
         let src = diagnostic.source_code();
         self.render_snippets(f, diagnostic, src)?;
@@ -190,6 +189,7 @@ impl GraphicalReportHandler {
             );
             write!(header, "{}", link)?;
             writeln!(f, "{}", header)?;
+            writeln!(f)?;
         } else if let Some(code) = diagnostic.code() {
             write!(header, "{}", code.style(severity_style),)?;
             if self.links == LinkStyle::Text && diagnostic.url().is_some() {
@@ -197,6 +197,7 @@ impl GraphicalReportHandler {
                 write!(header, " ({})", url.style(self.theme.styles.link))?;
             }
             writeln!(f, "{}", header)?;
+            writeln!(f)?;
         }
         Ok(())
     }
@@ -301,7 +302,6 @@ impl GraphicalReportHandler {
                     Some(Severity::Advice) => write!(f, "Advice: ")?,
                 };
                 self.render_header(f, rel)?;
-                writeln!(f)?;
                 self.render_causes(f, rel)?;
                 let src = rel.source_code().or(parent_src);
                 self.render_snippets(f, rel, src)?;
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -3,7 +3,7 @@ use std::fmt::{self, Write};
 use owo_colors::{OwoColorize, Style};
 use unicode_width::UnicodeWidthChar;
 
-use crate::diagnostic_chain::DiagnosticChain;
+use crate::diagnostic_chain::{DiagnosticChain, ErrorKind};
 use crate::handlers::theme::*;
 use crate::protocol::{Diagnostic, Severity};
 use crate::{LabeledSpan, MietteError, ReportHandler, SourceCode, SourceSpan, SpanContents};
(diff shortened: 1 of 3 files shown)


zkat (commented):
I'm a little concerned about all the extra whitespace, and the potential complexity of displaying multiple errors in their entirety (as opposed to just the header, which I think was ideal).

Any time we make changes like these, it's very important to understand what the end user experience might be like, both for developers and error consumers, and I'm not sure this matches up with what I'd like to see.

At the same time, there's a chance this is exactly the right way to address the "Related" errors issue that's been kinda hard to figure out.

I'd love to see more examples and in-depth discussion of the intention behind this design.

src/handlers/graphical.rs
```diff
@@ -231,7 +232,21 @@ impl GraphicalReportHandler {
...
+                match error {
+                    ErrorKind::Diagnostic(diag) => {
+                        let mut inner = String::new();
+
+                        // Don't print footer for inner errors
+                        let mut inner_renderer = self.clone();
+                        inner_renderer.footer = None;
+                        inner_renderer.render_report(&mut inner, diag)?;
+
+                        writeln!(f, "{}", textwrap::fill(&inner, opts))?;
```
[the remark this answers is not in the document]
TheNeikos: The idea is that if a developer chooses to go into giving a complete error experience they do not have to go and implement their own printer and instead can re-use the existing infrastructure.

As a concrete example I have the following error tree possible:

- User gives a configuration file with a variable amount of components and how to initialize them
    - Each component then gets loaded and can fail for various reasons, so that users can easily fix their errors all errors are collected and exposed as `related`. 
        - Each of these errors are a Diagnostic so that users get complete information with `source_code`s etc

An example output of this:

image

Now, this screenshot is without this patch, but once added I will be able to attach more info to the inner error and thus have a clear hierarchy of what happened and how the user could fix it.

TheNeikos:
> I'm a little concerned about all the extra whitespace, and the potential complexity of displaying multiple errors in their entirety (as opposed to just the header, which I think was ideal).

I am not sure why `miette` would need to be aware of this? If a user wishes to not have it displayed, they can choose not to use the features that `miette` gives them. Since with an empty impl of `Diagnostic` it looks exactly how the `Error` output.

>  I'm not sure this matches up with what I'd like to see.

I absolutely understand that this is what I'd call a 'controversial' change, so I def want this to evolve to help everyone using `miette`. Could you elaborate a bit more on what parts you think do not fit? (Unless they are those you've already mentioned)

I'd love to come up with some more examples/combinations of how this could look like in some situations. (I've also gone one step further regarding the `related` questions in #171 but I'd say that is even more controversial and can wait :P)

zkat:
>  If a user wishes to not have it displayed, they can choose not to use the features that miette gives them.

At scale, a dev does not necessarily have control over whether the sources they display are Diagnostics or not, and they might be surprised when they suddenly get a ton more output.

> Could you elaborate a bit more on what parts you think do not fit? (Unless they are those you've already mentioned)

I just think we can do better than just rendering the entire diagnostic inline like this.

> I'd love to come up with some more examples/combinations of how this could look like in some situations.

Ultimately, this is what I really want (for this and for #171). I'm not sure I like how they look right now, and I'd like to explore some more ideas of how things look _before_ we move on to implementation. Just figure out the best end result, regardless of current implementation, and work backwards from there. Heck, we can even change the general structure of error printing right now if it makes sense!

TheNeikos:
> At scale, a dev does not necessarily have control over whether the sources they display are Diagnostics or not, and they might be surprised when they suddenly get a ton more output.

Hm, I'd agree if `source` and `diagnostic_source` would be the same annotation, but if a dev puts `diagnostic_source` into their code then I do believe they should expect it to be printed as a diagnostic! (Otherwise, they could just use it as an `Error` & `source`)

> I just think we can do better than just rendering the entire diagnostic inline like this.
Alright! I am not exactly sure what the alternatives are haha, but I'd love to discuss it.

> Ultimately, this is what I really want (for this and for https://github.com/zkat/miette/pull/171). I'm not sure I like how they look right now, and I'd like to explore some more ideas of how things look before we move on to implementation. Just figure out the best end result, regardless of current implementation, and work backwards from there. Heck, we can even change the general structure of error printing right now if it makes sense!

Sounds good to me, I'll be able to get back to it probably next week, so I'll try to hash out some nice error structures then, so we can compare/discuss.

Porges:
> if a dev puts `diagnostic_source` into their code then I do believe they should expect it to be printed as a diagnostic!

Yes, definitely agree here. It’s opt-in and I would expect (as seen in #172) the diagnostics to be included in this case.

zkat:
Ok I think we can merge this. Do you mind rebasing/resolving conflicts?

TheNeikos:
@zkat there we go!

zkat (commented):
One last thing: do you mind writing a quick test to show what these nested diagnostics actually look like?

TheNeikos:
> One last thing: do you mind writing a quick test to show what these nested diagnostics actually look like?

Duh, totally forgot that one haha

It prints them indented as before, but simply prints them as a whole rather than just their debug info

tests/test_diagnostic_source_macro.rs
```diff
@@ -59,3 +74,120 @@ fn test_diagnostic_source() {
...
+        help: You should fix this
+      
+   ╭────
+ 1 │ right here
+   ·       ──┬─
+   ·         ╰── here
+   ╰────
+
+  Yooo, a footer
+"#;
```
TheNeikos, line 190: This is the crux of the change IMO. You can see that nested diagnostics actually get printed out as they are intended to be IMO

tests/test_diagnostic_source_macro.rs
```diff
@@ -59,3 +74,120 @@ fn test_diagnostic_source() {
...
+         ·    ──┬─
+         ·      ╰── here
+         ╰────
+        help: You should fix this
+      
+   ╭────
+ 1 │ right here
+   ·       ──┬─
+   ·         ╰── here
+   ╰────
```
TheNeikos, line 187: Currently causes come before snippets. This makes it look somewhat awkward. But I am not sure that this should be changed (if at all?) in this PR.

zkat (approved):

Porges:
Thanks all for getting this in! 🎉

PR #171, opened 2022-05-06
Group related diagnostics visually

TheNeikos:
I understand that this can be a controversial change, so this is very much a suggestion. The idea would be to group related errors so that end-user understands what exactly caused the problems they've encountered.

An example from the tests:
```
Error: oops::my::bad

  × oops!
   ╭─[bad_file.rs:1:1]
 1 │ source
 2 │   text
   ·   ──┬─
   ·     ╰── this bit here
 3 │     here
   ╰────
  help: try doing it better next time?
╭─There were 2 related diagnostics:
├─ 1.Error: oops::my::bad
│
│
│   × oops!
│    ╭─[bad_file.rs:1:1]
│  1 │ source
│    · ───┬──
│    ·    ╰── this bit here
│  2 │   text
│    ╰────
│   help: try doing it better next time?
│ ╭─There were 1 related diagnostics:
│ ├─ 1.Error: oops::my::bad
│ │
│ │
│ │   × oops!
│ │    ╭─[bad_file.rs:1:1]
│ │  1 │ source
│ │    · ───┬──
│ │    ·    ╰── this bit here
│ │  2 │   text
│ │    ╰────
│ │   help: try doing it better next time?
│ │ ├─There were 0 related diagnostics:
│ │
│
├─ 2.Error: oops::my::bad
│
│
│   × oops!
│    ╭─[bad_file.rs:1:1]
│  1 │ source
│    · ───┬──
│    ·    ╰── this bit here
│  2 │   text
│    ╰────
│   help: try doing it better next time?
│ ├─There were 0 related diagnostics:
│
```

### DIFF
--- a/tests/graphical.rs
+++ b/tests/graphical.rs
@@ -806,6 +806,7 @@ fn related() -> Result<(), MietteError> {
 
 Error: oops::my::bad
 
+
   × oops!
    ╭─[bad_file.rs:1:1]
  1 │ source
@@ -868,6 +869,7 @@ fn related_source_code_propagation() -> Result<(), MietteError> {
 
 Error: oops::my::bad
 
+
   × oops!
    ╭─[bad_file.rs:1:1]
  1 │ source
@@ -900,8 +902,7 @@ fn zero_length_eol_span() {
     let out = fmt_report(err.into());
     println!("Error: {}", out);
 
-    let expected = r#"
-  × oops!
+    let expected = r#"  × oops!
    ╭─[issue:1:1]
  1 │ this is the first line
  2 │ this is the second line
--- a/tests/graphical.rs
+++ b/tests/graphical.rs
@@ -5,6 +5,7 @@ use miette::{
     NarratableReportHandler, Report, SourceSpan,
 };
 use thiserror::Error;
+use pretty_assertions::assert_eq;
 
 fn fmt_report(diag: Report) -> String {
     let mut out = String::new();
@@ -784,11 +785,22 @@ fn related() -> Result<(), MietteError> {
     let err = MyBad {
         src: NamedSource::new("bad_file.rs", src.clone()),
         highlight: (9, 4).into(),
-        related: vec![MyBad {
-            src: NamedSource::new("bad_file.rs", src),
-            highlight: (0, 6).into(),
-            related: vec![],
-        }],
+        related: vec![
+            MyBad {
+                src: NamedSource::new("bad_file.rs", src.clone()),
+                highlight: (0, 6).into(),
+                related: vec![MyBad {
+                    src: NamedSource::new("bad_file.rs", src.clone()),
+                    highlight: (0, 6).into(),
+                    related: vec![],
+                }],
+            },
+            MyBad {
+                src: NamedSource::new("bad_file.rs", src.clone()),
+                highlight: (0, 6).into(),
+                related: vec![],
+            },
+        ],
     };
     let out = fmt_report(err.into());
     println!("Error: {}", out);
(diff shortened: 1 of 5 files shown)


tests/graphical.rs
```diff
@@ -803,18 +815,46 @@ fn related() -> Result<(), MietteError> {
...
+│ 
+│   × oops!
+│    ╭─[bad_file.rs:1:1]
+│  1 │ source
+│    · ───┬──
+│    ·    ╰── this bit here
+│  2 │   text
+│    ╰────
+│   help: try doing it better next time?
+│ ├─There were 0 related diagnostics:
```
LukeMathWalker, line 856: I'd expect this line to simply be omitted when there are no related diagnostics (what I expect to be the most common case to be fair).

tests/graphical.rs
```diff
@@ -803,18 +815,46 @@ fn related() -> Result<(), MietteError> {
...
-   · ───┬──
-   ·    ╰── this bit here
- 2 │   text
-   ╰────
-  help: try doing it better next time?
-
+╭─There were 2 related diagnostics:
+├─ 1.Error: oops::my::bad
+│ 
+│
```
LukeMathWalker, line 821: Is doubling the number of empty new lines deliberate or an oversight?
It does hurt the sense of visual grouping for me.

tests/graphical.rs
```diff
@@ -865,16 +905,18 @@ fn related_source_code_propagation() -> Result<(), MietteError> {
...
-Error: oops::my::bad
-
-  × oops!
-   ╭─[bad_file.rs:1:1]
- 1 │ source
-   · ───┬──
-   ·    ╰── this bit here
- 2 │   text
-   ╰────
+╭─There were 1 related diagnostics:
```
LukeMathWalker, line 908: The vertical `|` works quite nicely to "mark" that the entire block is composed of related diagnostics, but it does feel like it crowds the left area: I would find it more difficult to quickly scan the output and determine how many errors I encountered.

I wonder if this could be helped by nesting the vertical line itself a bit more - e.g. 
```
  There were 1 related diagnostics:
  │ 
  ├─ 1.Error: oops::my::bad
  │ 
  │   × oops!
  │    ╭─[bad_file.rs:1:1]
  │  1 │ source
  │    · ───┬──
  │    ·    ╰── this bit here
  │  2 │   text
  │    ╰────
```

kleinesfilmroellchen:
Several considerations from a user that would benefit greatly from this change:
- "There were 1 related diagnostics" => "There was 1 related diagnostic"
- "1.Error" => "1. Error"
- As @LukeMathWalker pointed out, remove the "related" lead line if there are no related errors.
- I still think this is too verbose for my use case. It may be worth the consideration of making the "related errors" style configurable for the GraphicalReportHandler and introduce some other styles. I'm especially thinking of an alternate, fully inline style (like what is currently done, but without just listing the different errors).

CyriacBr:
I also think it'd be nice to be able group related errors better. The way related errors are displayed right now, each error seem completely independent, which make them actually seem unrelated.  
Here's (very) roughly what one of my use case look like:  
```bash
  × A runtime error occurred

Error: 
  × Cannot add a number to a bool
   ╭─[5:1]
 5 │             fn b($n)
 6 │                 $n + a()
   ·                 ─┬ ┬ ─┬─
   ·                  │ │  ╰── number
   ·                  │ ╰── invalid operation
   ·                  ╰── bool
 7 │             end
   ╰────
Error: code

  × error originated from here
    ╭─[8:1]
  8 │             fn c($n)
  9 │                 b($n)
    ·                 ──┬──
    ·                   ╰── error originated here
 10 │             end
    ╰────
Error: code

  × error originated from here
    ╭─[10:1]
 10 │             end
 11 │             c(true)
    ·             ───┬───
    ·                ╰── error originated here
 12 │             
    ╰────
```

I see that it's still unsure what direction to take regarding this issue.
I find the current proposal rather verbose, and in my case I think it'd look much better just without the `Error: ...` clause. If there was an easy way to remove them, it'd make related errors more... related.

TheNeikos:
I don't think this is super salvageable anymore, but if anyone wants to pick this up, feel free to!

PR #173, opened 2022-05-13
Fix asref diagnostic macro call

### LINKED ISSUE #172: Wrapping other Errors without losing diagnostics
I’m having trouble figuring out how to directly (i.e. not with a set of errors like `related`) wrap another Miette-embellished error type without losing its diagnostic information.

I started with (for example):

```rs
#[derive(Error, Diagnostic, Debug)]
pub enum Error {
    // …
    #[error("Input file had a syntax error")]
    #[diagnostic(code(mylib::syntax_error))]
    SyntaxErr(#[from] knuffel::Error),
}
```

However, when printing this drops the diagnostic information from the `knuffel::Error`.

It seems like `source`/`diagnostic_source` are the solution to this (and `from` implies `source`), however if I use them like this:

```rs
#[derive(Error, Diagnostic, Debug)]
pub enum Error {
    // …
    #[error("Input file had a syntax error")]
    #[diagnostic(code(mylib::syntax_error))]
    SyntaxErr{
        #[from]
        #[source]
        #[diagnostic_source]
        source: knuffel::Error,
    },
}
```

… I get problems with the `Diagnostic` macro:

```
error[E0599]: the method `as_ref` exists for reference `&knuffel::Error`, but its trait bounds were not satisfied
  --> src/lib.rs:8:17
   |
8  | #[derive(Error, Diagnostic, Debug)]
   |                 ^^^^^^^^^^ method cannot be called on `&knuffel::Error` due to unsatisfied trait bounds
   |
  ::: …/knuffel-2.0.0/src/errors.rs:27:1
   |
27 | pub struct Error {
   | ---------------- doesn't satisfy `knuffel::Error: AsRef<_>`
   |
   = note: the following trait bounds were not satisfied:
           `knuffel::Error: AsRef<_>`
           which is required by `&knuffel::Error: AsRef<_>`
   = note: this error originates in the derive macro `Diagnostic` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
```

`#[diagnostic_source]` actually doesn’t fail if I lift it onto the enum variant (as opposed to its field) but then it doesn’t seem to have any effect.

I feel like this should be something straightforward, what am I missing? 😊

matthiasbeyer:
Closes #172 

I would like to ask @TheNeikos for review as well here, as #165 was developed with them.

### DIFF
--- a/miette-derive/src/diagnostic_source.rs
+++ b/miette-derive/src/diagnostic_source.rs
@@ -59,7 +59,7 @@ impl DiagnosticSource {
                     };
                     quote! {
                         Self::#ident #display_pat => {
-                            std::option::Option::Some(#rel.as_ref())
+                            std::option::Option::Some(#rel)
                         }
                     }
                 })
--- a/tests/test_diagnostic_source_macro.rs
+++ b/tests/test_diagnostic_source_macro.rs
@@ -18,3 +18,22 @@ fn test_diagnostic_source() {
     };
     assert!(error.diagnostic_source().is_some());
 }
+
+// Compiletest this:
+#[derive(thiserror::Error, miette::Diagnostic, Debug)]
+enum EnumError {
+    #[error("Some Error text")]
+    SyntaxErr(#[from] AnErr),
+}
+
+// Compiletest this:
+#[derive(thiserror::Error, miette::Diagnostic, Debug)]
+enum AnotherEnumError {
+    #[error("Some other Error Text")]
+    SyntaxErr {
+        #[from]
+        #[source]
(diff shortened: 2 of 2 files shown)


matthiasbeyer:
This is done way better with #169 IMO. I guess this one should be closed in favour of #169!

matthiasbeyer:
Maybe 0b32574 can be added to #169, so the error from #172 is covered there explicitely as well.

zkat:
Closing in favor of #169 which I just merged :)

Issue #176, opened 2022-05-18
Default to ANSI colors even if truecolor is available?


Benjamin-L:
Currently the default color scheme from `GraphicalTheme::default` will prefer truecolor to ansi colors, if available. Many users have terminals with a specific color scheme configured for the ANSI colors, and CLI tools that synthesize their own color scheme with arbitrary RGB values tend to look very out of place. This is particularly bad with a non-black background, which can result in text with unreadably low contrast. My personal preference would be for truecolor to only be used by default in situations where the exact color value matters rather than the semantic label of the color. For example, truecolor-by-default makes sense for a CLI image viewer or a tool to visualize color hex codes. Error diagnostics fall pretty cleanly into the "exact color values don't matter but semantics do" group.

This is, however, a subjective thing, and no hard feelings if the `miette` maintainers don't agree and want to keep the current behavior. I've set up `FORCECOLOR=1` on my machine, so things look fine for me either way.

sunshowers:
Just wanted to say thanks for doing this @Benjamin-L! I did indeed raise this point in #150 as well but never had the time to follow up.

PR #177, opened 2022-05-25
feat(theme): restructure automatic color selection

Benjamin-L:
Fixes: #176.

The original discussion in #176 was about changing the default from RGB to ANSI, however once I started work on that I ended up restructuring the options to control automatic color selection, in order to allow a wider range of behaviors and decouple the variables a bit. This is a much larger breaking change, so I'd be happy to pull that out if we don't want to deal with the API change. I'm also a little unsure on the value of the unit tests for this, given that I have ~130 lines of testing for a relatively straightforward if-statement.

BREAKING CHANGES:
* the default theme now prefers ANSI colors, even if RGB is supported
* `MietteHandlerOpts::ansi_colors` is removed
* `MietteHandlerOpts::rgb_color` now takes an enum that controls the
  color format used when color support is enabled, and has no effect
  otherwise.

### DIFF
--- a/tests/color_format.rs
+++ b/tests/color_format.rs
@@ -1,0 +1,137 @@
+#![cfg(feature = "fancy-no-backtrace")]
+
+use lazy_static::lazy_static;
+use miette::{Diagnostic, MietteHandler, MietteHandlerOpts, ReportHandler, RgbColors};
+use regex::Regex;
+use std::fmt::{self, Debug};
+use std::sync::Mutex;
+use thiserror::Error;
+
+#[derive(Eq, PartialEq, Debug)]
+enum ColorFormat {
+    NoColor,
+    Ansi,
+    Rgb,
+}
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("oops!")]
+#[diagnostic(code(oops::my::bad), help("try doing it better next time?"))]
+struct MyBad;
+
+struct FormatTester(MietteHandler);
+
+impl Debug for FormatTester {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.0.debug(&MyBad, f)
+    }
+}
+
+/// Check the color format used by a handler.
+fn color_format(handler: MietteHandler) -> ColorFormat {
+    let out = format!("{:?}", FormatTester(handler));
+
+    let rgb_colors = Regex::new(r"\u{1b}\[[34]8;2;").unwrap();
+    let ansi_colors = Regex::new(r"\u{1b}\[(3|4|9|10)[0-7][m;]").unwrap();
+    if rgb_colors.is_match(&out) {
+        ColorFormat::Rgb
+    } else if ansi_colors.is_match(&out) {
+        ColorFormat::Ansi
+    } else {
+        ColorFormat::NoColor
+    }
+}
+
+/// Runs a function with an environment variable set to a specific value, then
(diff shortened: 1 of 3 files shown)


Benjamin-L:
I've added another commit to change the default ANSI theme to make it (subjectively) better on different terminal color schemes, which is something we discussed in #176. Whether or not this is actually better is obviously up to interpretation, but my main issue with the original theme was that the first span was colored red, which was the same color used for errors. The colors I landed on for the spans are *relatively* similar to the ones used by rgb on the more "normal" color schemes I tried.

**RGB theme:**
a screenshot of an example error message using RGB colors with various terminal color schemes

**Original ANSI theme:**
a screenshot of an example error message using the original ANSI colors with various terminal color schemes

**New ANSI theme:**
a screenshot of an example error message using the new ANSI colors with various terminal color schemes

zkat:
wow these are so much more readable. Thank you for doing this! That's a lot of work!

zkat:
(I haven't forgotten about this, I'm just debating whether I should just merge this now, or wait for a bit before doing a full semver-major for miette again. I should probably just bite the bullet.)

Benjamin-L:
I'm good either way. Bumping the major version immediately after a "start the next development iteration with a minor version bump" commit definitely feels a little silly :)

A couple questions about this PR after looking it over again:

Should the dev-dependency additions be split out into their own `dep(...)` commits, or is that just for normal dependencies?

I think this crate is targeting windows support, and am not 100% sure whether the tests I added here will run correctly on windows. I honestly don't have any experience with how windows stuff handles colors. My understanding is that some (but not all) windows terminals have out-of-band color formatting with actual syscalls for color changes or something like that. My fear is that either the `supports_color` crate is going to give different results on certain windows terminals (even with the environment variable shenanigans), or that the color formatting is going to output something other than ANSI escape codes on a windows machine, which would break the `color_format` function in the test. It looks like there is a CI run for windows, but I'm not sure if that environment would actually trigger any issues here. I don't have a windows machine to test with, unfortunately, but if you've got a better understanding of how this works and/or a machine to test with, that would be ideal. I can also do some more research.

tests/color_format.rs
```diff
@@ -0,0 +1,137 @@
...
+    Ansi,
+    Rgb,
+}
+
+#[derive(Debug, Diagnostic, Error)]
+#[error("oops!")]
+#[diagnostic(code(oops::my::bad), help("try doing it better next time?"))]
+struct MyBad;
+
+struct FormatTester(MietteHandler);
```
Benjamin-L, line 22: Is there a better way to do this? Not super familiar with miette's API, and there might be something I missed.

zkat: https://github.com/zkat/miette/blob/main/tests/graphical.rs#L9-L35 this is how the current tests do this, if I understand your question.

zkat:
https://github.com/zkat/miette/runs/6604788328?check_suite_focus=true The tests are already running on Windows :)

tests/color_format.rs
Benjamin-L: Yeah, that would be the equivalent for the other tests. Unfortunately, calling `render_report` directly doesn't work here because the automatic color selection happens as part of `MietteHandler`, which just implements the `ReportHandler` trait, and doesn't have an equivalent to `render_report`. Gonna leave the `FormatTester` struct as-is.

PR #180, opened 2022-06-24
Don't escape single-quotes, that's not valid json

Gankra:
See for instance how this json parser gets mad:

```json
{
  "hello": "\\ \" \n \r \t \b \f"
  "oops": "this is evil \' oh no!"
}
```


Gankra:
I'm also a little suspicious of this line, which doesn't escape the URL, but I haven't looked to see if the url printing is pre-escaped (sounds dubious...).

https://github.com/zkat/miette/blob/714334098a92c77fb6b962e627defab7e16b540d/src/handlers/json.rs#L84

PR #181, opened 2022-06-25
Implement SourceCode for Arc<str> (and similar types)

andrewhickman:
Adding the ?Sized bound allows use of Arc<str> as a source code type

### DIFF
--- a/src/source_impls.rs
+++ b/src/source_impls.rs
@@ -158,7 +158,7 @@ impl SourceCode for String {
     }
 }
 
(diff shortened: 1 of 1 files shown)


zkat:
Just to make sure I understand because I haven't played with `?Sized` a lot--this shouldn't be a breaking change, right? It only allows _more_ things to be used?

andrewhickman:
Yes, that's right.

zkat (approved):

zkat:
released as part of miette@5.1.0. Thanks!

PR #183, opened 2022-07-09
fix(graphical): handle an empty source

sunshowers:
In some cases the source can be completely empty -- handle that in a reasonable fashion.


zkat (approved):

zkat:
Thanks again! This has been tagged and released as part of miette 5.1.1 :)

Issue #191, opened 2022-08-07
Allow disabling the rendering of the cause chain


LukeMathWalker:
Hi!
I have been using `miette` for a compiler-like project recently - I have been loving it so far, but I am facing a dilemma.

I'd love to have the fancy-like features for the user-facing reports, but I don't want to expose my users to the cause chain of the top-level error - it would leak implementation details that they should not care about.
As far as I can see, there is no way to suppress the rendering of the cause chain; I can either render the top-level error message (with no source/labels/helps) via `Display` or I can get the rich error but I end up leaking lower-level error details via `Debug`.

Would adding a new knob to [`MietteHandlerOpts`](https://docs.rs/miette/latest/miette/struct.MietteHandlerOpts.html) be a welcomed addition?

LukeMathWalker:
It turns out it was quite straight-forward to sketch it out, so I submitted it a PR. Feel free to reject it if you don't think this is a desired configuration option.

PR #192, opened 2022-08-07
Allow miette users to opt-out of the rendering of the cause chain

LukeMathWalker:
Related to #191 

The default behaviour remains to render the cause chain, but knobs are exposed to allow avoiding cause chain rendering.


zkat (approved):
This looks great! Thanks for the contribution!

zkat:
Looks like you need to fix formatting? This otherwise looks fine to me :)

LukeMathWalker:
My bad! It should be fixed now 😁

PR #202, opened 2022-09-10
Fix graphical label alignment issues 

### LINKED ISSUE #87: Indentation with tabs throws off labels
(Sorry to not provide a proper repro. I’ll provide one later if you want, just wanted to get the issue opened before I forget).

If I have code indented with tabs, it seems that something goes wrong with the labelled spans:
```ts
function lol(): void {
	const x: i32 = 1;
};
```

Screenshot 2021-10-10 at 14 35 29

If I instead indent the code with spaces, it looks correct:

Screenshot 2021-10-10 at 14 35 35

I’m assuming that this might be due to terminal configuration how to display tabs?

Benjamin-L:
Fixes #87 and #97.

Turns out that reliably determining the width of a string when displayed in a terminal is impossible without using ANSI escapes to get the current cursor position. Relying on the ANSI escapes is an option, but won't work if your output isn't a TTY. Supporting non-TTY output is important for things like redirecting error messages to a file, so I implemented a _mostly correct_ label alignment process that makes some specific tradeoffs. In the future, we could also add the escape-based option, and just fall back to this behavior on non-TTY output.

Back in https://github.com/zkat/miette/issues/87#issuecomment-1137930566, I had an idea for a hack to fix tab alignment while still respecting the terminal's tabstop setting. Unfortunately, I couldn't figure out a way to make this work for underlines, so the remaining option is to give up and replace tabs with spaces. This allows us to determine the visual width of the output reliably, but it does mean that we're ignoring the user's tabstop preferences. The original default behavior, where tabs were emitted verbatim and highlight alignment was broken, is pretty much never desirable, so I ended up just removing that option.

One of the issues with replacing tabs with a fixed number of spaces is that it will produce bad results if tabs are used for alignment in the middle of a line. To avoid this, we can emulate the behavior of the terminal, where tabs shift the cursor to the next tabstop rather than by a fixed width. Existing compilers that I checked seem to go both ways on this one. Gcc, clang, and rustc all emulate tabstops, while ghc uses a fixed width. I implemented the tabstop emulation option, which is a second breaking change, but this should be easy enough to swap out if we don't want it.

The other tradeoff here is that `unicode_width` isn't actually going to be able to predict the width of characters displayed on an arbitrary terminal. For example, 🏳️‍🌈 could show up as 1, 2, 3, or 4 columns. Predicting this is, in the general case, not possible. `unicode_width` is still better than assuming every character is 1 column though, so yay for incremental improvement.

Computers are a mess lol.

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -1,7 +1,7 @@
 use std::fmt::{self, Write};
 
 use owo_colors::{OwoColorize, Style};
-use unicode_width::UnicodeWidthStr;
+use unicode_width::UnicodeWidthChar;
 
 use crate::diagnostic_chain::DiagnosticChain;
 use crate::handlers::theme::*;
@@ -28,7 +28,7 @@ pub struct GraphicalReportHandler {
     pub(crate) theme: GraphicalTheme,
     pub(crate) footer: Option<String>,
     pub(crate) context_lines: usize,
-    pub(crate) tab_width: Option<usize>,
+    pub(crate) tab_width: usize,
     pub(crate) with_cause_chain: bool,
 }
 
@@ -49,7 +49,7 @@ impl GraphicalReportHandler {
             theme: GraphicalTheme::default(),
             footer: None,
             context_lines: 1,
-            tab_width: None,
+            tab_width: 4,
             with_cause_chain: true,
         }
     }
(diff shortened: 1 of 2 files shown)


zkat (approved):
This is a great PR! Thanks so much for taking the time to do this!

I think you made the right compromises here and I think it's probably overkill to start trying to understand terminals themselves. Maybe some day, but not now :P

I'm really happy to see this so significantly improved! I'll push out a release with this soon.

zkat:
Note: I don't consider this a breaking change because _rendering_ isn't part of the API contract for miette, so this is gonna just go out as a patch release.

PR #216, opened 2022-10-30
Implement `SourceCode` for `Vec<u8>`

Phantomical:
SourceCode is implemented for &str, String, and &[u8] but it is missing an impl for `Vec<u8>`.

### DIFF
--- a/src/source_impls.rs
+++ b/src/source_impls.rs
@@ -119,6 +119,17 @@ impl<'src> SourceCode for &'src [u8] {
     }
 }
 
+impl SourceCode for Vec<u8> {
+    fn read_span<'a>(
+        &'a self,
+        span: &SourceSpan,
+        context_lines_before: usize,
+        context_lines_after: usize,
+    ) -> Result<Box<dyn SpanContents<'a> + 'a>, MietteError> {
+        <[u8] as SourceCode>::read_span(self, span, context_lines_before, context_lines_after)
+    }
+}
+
 impl SourceCode for str {
     fn read_span<'a>(
         &'a self,


zkat (approved):

Benjamin-L:
Thoughts on just replacing the existing impls with the following?

```rust
impl SourceCode for [u8] { ... }
impl SourceCode for str { ... }
impl<T: Deref + Send + Sync> SourceCode for T  where T::Target: SourceCode { ... }
```

This should cover all the existing impls, `Vec<u8>`, and any other smart pointer types, including ones outside of `std`.

zkat:
wait really?

If that won't be a breaking change, that does seem better!

Benjamin-L:
It's _technically_ a breaking change, because it may overlap with existing downstream impls. For example, suppose the following code was in some dependent crate:

```rust
struct Foo(String);

impl Deref for Foo {
    type Target = str;
    ...
}

impl SourceCode for Foo { ... }
```

That `SourceCode` impl is now invalid, because it conflicts with the blanket impl we're adding, breaking the trait coherence rules.

RFC 1105 states that implementing traits [is only considered minor](https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-implementing-any-non-fundamental-trait) though, even though it is always technically breaking. However, the type of breaking change discussed there has to do with dispatch ambiguity rather than coherence, so I'm not sure the same reasoning applies. Fixing dispatch ambiguity issues is very straightforward, while fixing this type of coherence issue might require much deeper changes.

I'm not sure how much of an issue this is in practice. I wish there was an easy way to just build all the dependent crates against a proposed change like this, similar to how crater works for rustc changes :)

Phantomical:
Would it be alright to merge this PR by itself and then have a second one with the breaking impl based on `Deref`? The changes here right now are not breaking and it would be nice to have the impl for `Vec<u8>` available.

Benjamin-L:
Yeah, when I originally suggested the `Deref` option I hadn't thought through the fact that it's potentially breaking.

zkat:
Let's merge this in for now, then :)

Issue #215, opened 2022-10-29
Panic when span is on `\n` in v5.4.1


Boshen:
Input text: "/\n"
Input span: 0..2

thread '\<unnamed\>' panicked at 'byte index 2 is out of bounds of `/`', /Users/boshenchen/.cargo/registry/src/github.com-1ecc6299db9ec823/miette-5.4.1/src/handlers/graphical.rs:624:21

https://github.com/zkat/miette/blob/3e25fd5b86324e16b33dd7e9192faa8015ead562/src/handlers/graphical.rs#L624

cc @Benjamin-L

Boshen:
The previous version was able to point to the newline.

```
  × Unterminated regular expression
   ╭─[./target/main.ts:1:1]
 1 │ /
   ·  ┬
   ·  ╰── 
   ╰────
```

Boshen:
It also panics if the span is on EOF.

zkat:
@Benjamin-L lol are we still being cursed by the same issue?

zh217:
I also had this problem on 5.4.1 and 5.4.0. Version 5.3.0 is fine.

Benjamin-L:
Oof, yeah I didn't realize that spans that contained the EOL were previously supported. I'll have a fix for this up in a bit. While working this out, I found a few edge cases that we'll want to make sure we have good behavior for:

### Edge Cases

#### CRLFs

In v5.3.0, these end up rendering like this:

Source: `source\r\ntext\r\n  here`
Span: `(8, 6)`

```
Error: oops::my::bad

  × oops!
   ╭─[bad_file.rs:1:1]
 1 │ source
 2 │ text
   · ───┬──
   ·    ╰── this bit here
 3 │   here
   ╰────
  help: try doing it better next time?
```

This seems... probably not desirable, and instead the rendered highlight should only extend one visual column past the end of line.

#### EOF

Versions past 5.4.0 panic on a span including EOF. v5.3.0 doesn't panic, but the output is bad:

Source: `source\ntext`
Span: `(7, 5)`

```
Error: oops::my::bad

  × oops!
   ╭─[bad_file.rs:1:1]
 1 │     source
 2 │ ╭─▶ text
   ╰────
  help: try doing it better next time?
```

Similar to the CRLF case, I think the rendered highlight should only extend one column past the end of line.

Benjamin-L:
> It also panics if the span is on EOF.

[...]

PR #221, opened 2022-11-03
fix(graphical): Fix panic with span extending past end of line (#215)

Benjamin-L:
This also changes the behavior with spans including a CRLF line-ending. Before the panic bug was introduced, these were rendered with the CRLF being two visual columns wide. Now, any span extending past the EOL is treated as including one extra visual column. See https://github.com/zkat/miette/issues/215#issuecomment-1301560478 for an example of this.

This PR notably does not fix the rendering bug with spans including the EOF. These were rendered incorrectly before the panic bug was introduced. @Boshen mentioned that they were seeing a panic with EOF in #215, but I haven't been able to reproduce that. We may want to wait for a response from them before merging this, since it seems like I'm missing something here.

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -621,8 +621,22 @@ impl GraphicalReportHandler {
         let line_range = line.offset..=(line.offset + line.length);
         assert!(line_range.contains(&offset));
 
-        let text = &line.text[..offset - line.offset];
-        self.line_visual_char_width(text).sum()
+        let text_index = offset - line.offset;
+        let text = &line.text[..text_index.min(line.text.len())];
+        let text_width = self.line_visual_char_width(text).sum();
+        if text_index > line.text.len() {
+            // Spans extending past the end of the line are always rendered as
+            // one column past the end of the visible line.
+            //
+            // This doesn't necessarily correspond to a specific byte-offset,
+            // since a span extending past the end of the line could contain:
+            //  - an actual \n character (1 byte)
+            //  - a CRLF (2 bytes)
+            //  - EOF (0 bytes)
+            text_width + 1
+        } else {
+            text_width
+        }
     }
 
     /// Renders a line to the output formatter, replacing tabs with spaces.
--- a/tests/graphical.rs
+++ b/tests/graphical.rs
@@ -359,6 +359,80 @@ fn single_line_higlight_offset_end_of_line() -> Result<(), MietteError> {
 }
 
 #[test]
+fn single_line_higlight_include_end_of_line() -> Result<(), MietteError> {
+    #[derive(Debug, Diagnostic, Error)]
+    #[error("oops!")]
+    #[diagnostic(code(oops::my::bad), help("try doing it better next time?"))]
+    struct MyBad {
+        #[source_code]
+        src: NamedSource,
+        #[label("this bit here")]
+        highlight: SourceSpan,
+    }
+
+    let src = "source\n  text\n    here".to_string();
+    let err = MyBad {
+        src: NamedSource::new("bad_file.rs", src),
+        highlight: (9, 5).into(),
+    };
+    let out = fmt_report(err.into());
+    println!("Error: {}", out);
+    let expected = r#"oops::my::bad
+
+  × oops!
+   ╭─[bad_file.rs:1:1]
+ 1 │ source
+ 2 │   text
+   ·   ──┬──
+   ·     ╰── this bit here
+ 3 │     here
+   ╰────
+  help: try doing it better next time?
+"#
+    .trim_start()
+    .to_string();
+    assert_eq!(expected, out);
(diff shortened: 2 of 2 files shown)


Boshen:
Yeah it was probably the bad behaviour.

To make sure everything is correct, I've tested your patch, a few points:
* It seems the graphic handler need to be updated as well, it crashed in the same spot. https://github.com/zkat/miette/blob/3e25fd5b86324e16b33dd7e9192faa8015ead562/src/handlers/graphical.rs#L620-L626

Boshen:
With the upgrade, I found another interesting case where I was pointing the span inside a unicode.


v5.3.0:

offset 5..7

```
  × Invalid Character `ⸯ`
    ╭─[language/identifiers/vertical-tilde-continue.js:15:1]
 15 │ 
 16 │ var aⸯ; // U+2E2F
    ·      ─┬
    ·       ╰── Invalid Character `ⸯ`
    ╰────
```

v5.4.0:
Span: 5..7 gives

'byte index 7 is not a char boundary; it is inside 'ⸯ' (bytes 5..8) of `var aⸯ;

I had to change the span to 5..8 to cover the unicode
```
  × Invalid Character `ⸯ`
   ╭─[main.ts:1:6]
 1 │ var aⸯ;
   ·      ┬
   ·      ╰── Invalid Character `ⸯ`
 2 │
   ╰────
```

Boshen:
And to make things more annoying ...

Although v5.4.0 is more accurate when pointing to a unicode, but the graphical reporter will use the range pointer `┬`, it would be even nicer to get it to use the single char pointer `▲`.

Benjamin-L:
> Yeah it was probably the bad behaviour.
> 
> To make sure everything is correct, I've tested your patch. It seems the graphic handler need to be updated as well, it crashed in the same spot.
> 
> https://github.com/zkat/miette/blob/3e25fd5b86324e16b33dd7e9192faa8015ead562/src/handlers/graphical.rs#L620-L626

To be clear, you tested input similar to the examples from #215 against this PR and are still seeing a panic? If so, could you send the input that you're seeing the panic on?

The code block linked in that comment is the same one that this PR is changing:

https://github.com/zkat/miette/pull/221/files#diff-a775e9b52316667cfcabc8d12a083cba8d18759534d48b7a5da6cd647bc6fe6dR620-R639

I tested the input from #215 during development of this. I was able to reproduce the panic with the `main` branch, but did not reproduce the panic with the code from this PR.

Benjamin-L:
> With the upgrade, I found another interesting case where I was pointing the span inside a unicode.
> 
> v5.3.0:
> 
> offset 5..7
> 
> ```
>   × Invalid Character `ⸯ`
>     ╭─[language/identifiers/vertical-tilde-continue.js:15:1]
>  15 │ 
>  16 │ var aⸯ; // U+2E2F
>     ·      ─┬
>     ·       ╰── Invalid Character `ⸯ`
>     ╰────
> ```
> 
> v5.4.0: Span: 5..7 gives
> 
> 'byte index 7 is not a char boundary; it is inside 'ⸯ' (bytes 5..8) of `var aⸯ;
> 
> I had to change the span to 5..8 to cover the unicode
> 
> ```
>   × Invalid Character `ⸯ`
>    ╭─[main.ts:1:6]
>  1 │ var aⸯ;
>    ·      ┬
>    ·      ╰── Invalid Character `ⸯ`
>  2 │
>    ╰────
> ```

I've created a separate issue for this: #223.

> And to make things more annoying ...
> 
> Although v5.4.0 is more accurate when pointing to a unicode, but the graphical reporter will use the range pointer `┬`, it would be even nicer to get it to use the single char pointer `▲`.

I don't think this is specific to multi-byte unicode characters. Miette currently only uses the `▲` pointer when the span has length 0 (it points to a specific index, but doesn't include the following character in the span). The `┬` pointer is used for any single-line span with nonzero length. This includes multi-byte characters that take a single visual column (length > 1), but also includes single-byte characters with a single visual column (length = 1).

IMO, it could go either way as to whether or not this behavior is correct. The one thing that I think is definitely _not_ correct is that multi-byte spans with zero visual columns (for example, spanning a zero-width-space or the `\r` of a CRLF with the code in this PR) are rendered as `┬` rather than `▲`, which is actively confusing. I'll make an issue for this one later today.

Benjamin-L:
@Boshen I created an issue for the `┬` usage on spans pointing at zero-width nonempty spans: #224. That issue is _probably_ the right place to discuss whether or not `▲` should be used for 1-wide spans.

Boshen:
> The code block linked in that comment is the same one that this PR is changing:

Oh I was looking at the wrong place again, I got dizzy from testing all these difference branches and versions 😵 

Anyway, this PR passed all my tests (I have a JavaScript / TypeScript compiler running with all Test262 / Babel / TypeScript conformance tests).

PR #233, opened 2023-01-29
Mention miette::miette! macro under "... in application code"

rukai:
Being able to construct ad-hoc errors is an important part of writing applications so I feel `miette::miette!` deserves to take up some space under this heading


zkat (approved):

zkat:
thanks!

PR #234, opened 2023-02-07
Fix wrong severity of related errors

### LINKED ISSUE #232: Related reports render with the severity of the parent report.
I would expect the related reports to have prefixed "Error: ", "Warning: " or "Advice: " according to the severity level they have, but instead it is printed with the severity level of the parent report.

Is this the intended behaviour or is it a bug?

---
With this code
```rust
#[derive(Diagnostic, Debug, Error)]
#[error("multiple errors")]
#[diagnostic(severity(error))]
pub struct MutiError {
    #[related]
    related: Vec<InnerError>,
}

#[derive(Diagnostic, Debug, Error)]
pub enum InnerError {
    #[error("a warning")]
    #[diagnostic(severity(warning))]
    Warning,
    #[error("an error")]
    #[diagnostic(severity(error))]
    Error,
}

fn main() -> miette::Result<()> {
    Err(MutiError {
        related: vec![InnerError::Warning, InnerError::Error],
    })?
}
```

I get this output:
image

But I would expect this:
image

---

I think is just these lines in the graphical and narratable reporters, where `diagnostic` is used instead of `rel`.

https://github.com/zkat/miette/blob/14f952dc91b9f27b7ae1e9e03b89744bcc682fa2/src/handlers/graphical.rs#L283-L288
https://github.com/zkat/miette/blob/14f952dc91b9f27b7ae1e9e03b89744bcc682fa2/src/handlers/narratable.rs#L134-L139

dalance:
Currently, the severity of the parent error is shown as the severity of related errors.
This PR fix it.

Closes #232

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -281,7 +281,7 @@ impl GraphicalReportHandler {
         if let Some(related) = diagnostic.related() {
             writeln!(f)?;
             for rel in related {
-                match diagnostic.severity() {
+                match rel.severity() {
                     Some(Severity::Error) | None => write!(f, "Error: ")?,
                     Some(Severity::Warning) => write!(f, "Warning: ")?,
                     Some(Severity::Advice) => write!(f, "Advice: ")?,
--- a/src/handlers/narratable.rs
+++ b/src/handlers/narratable.rs
@@ -132,7 +132,7 @@ impl NarratableReportHandler {
         if let Some(related) = diagnostic.related() {
             writeln!(f)?;
             for rel in related {
-                match diagnostic.severity() {
+                match rel.severity() {
                     Some(Severity::Error) | None => write!(f, "Error: ")?,
                     Some(Severity::Warning) => write!(f, "Warning: ")?,
                     Some(Severity::Advice) => write!(f, "Advice: ")?,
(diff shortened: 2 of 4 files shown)


zkat (approved):
oooh thanks for this!

zkat:
looks like you just need some clippy fixes

dalance:
Some clippy issues can't pass both 1.56.0 and stable.

* 1.56.0

```
error: there is no argument named `enum_name`
  --> miette-derive/src/url.rs:87:55
   |
87 |                         let item_path = format!("enum.{enum_name}.html#variant.{ident}");
   |
```

* stable

```
error: variables can be used directly in the `format!` string
  --> miette-derive/src/url.rs:87:41
   |
87 |                         let item_path = format!("enum.{}.html#variant.{}", enum_name, ident);
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
   = note: `-D clippy::uninlined-format-args` implied by `-D warnings
```

zkat:
Ah that's annoying. Let me look into making it so clippy only runs on stable, I guess. No point in running it on both anyway.

dalance:
If you want to use 1.56.0 as MSRV, clippy should do at 1.56.0 only because new features which cause error at stable's clippy can't be used.

zkat:
yeah, I guess you're right :)

zkat:
https://github.com/zkat/miette/commit/b658fc020b23b0715339c5c60f7c12c947f9a747 should fix this hopefully

zkat:
Looks like we're good now! Thanks again :)

PR #235, opened 2023-02-09
SourceCode provides access to bytes

Kylebrown9:
I would like to be able to use `&NamedSource` as the input to parsing and lexing functions that use and borrow from the source already stored in the `NamedSource::inner()` instead of passing another copy. This could simplify the API for a basic parse function to something like this:

```rust
pub fn parse(source: &NamedSource) -> Result<Tokens<'_>> {
   let chars = source.inner().source_str()?.char_indices();
   ...
}
```

This PR adds the `SourceCode::source_bytes()` and `SourceCode::source_str()` functions to enable this use case.

### DIFF
--- a/src/named_source.rs
+++ b/src/named_source.rs
@@ -53,4 +53,8 @@ impl SourceCode for NamedSource {
             contents.line_count(),
         )))
     }
+
+    fn source_bytes(&self) -> &[u8] {
+        self.inner().source_bytes()
+    }
 }
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -198,7 +198,7 @@ pub trait SourceCode: Send + Sync {
 
     /// Borrow the source bytes as a string slice.
     /// Returns None if the source is not a valid utf-8 string.
-    fn source_utf8(&self) -> Result<&str, std::str::Utf8Error> {
+    fn source_str(&self) -> Result<&str, std::str::Utf8Error> {
         std::str::from_utf8(self.source_bytes())
     }
 }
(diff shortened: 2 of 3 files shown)


zkat:
I'm not sure I want this change, tbh.

For one, it's a pretty significant breaking change to add two required trait methods.

For two, I feel like this API is kind of a footgun: while you can technically read an _entire_ file using `read_span`, miette was designed to work on very very large sources, much larger than you might want to fit in memory. For those very large sources, it would be pretty hard to provide a borrow-based API like this one, since they wouldn't want to keep their entire source in memory, even if someone decides they do want to slurp up an entire massive file.

I feel like a much better way to fulfill your specific ask for your API is to just implement an equivalent of NamedSource.

Alternatively, I wonder if what I should've done from the beginning was to make `NamedSource` be `NamedSource<T>` where `T: SourceCode + 'static`. I think that would work, no? And there's a better transition path for that one, by using a different name and deprecating `NamedSource` itself.

what do you think?

Kylebrown9:
Totally understand, I wasn't sure if you'd want it but I figured it was easy enough to sketch out and present as an option. The last option you suggest sounds reasonable to me. I'd be able to pick a value of `T` that when accessed through `NamedSource::inner()` would let me do the tokenizing and parsing I want.

I'm not sure exactly how/when you'd want to do that, but I'd be happy to create an issue for it and close this PR.

zkat:
Yeah let's make an issue for now describing the problem and linking back here and we can talk more about it there? I think this pr is just not it although I appreciate the time and I can see how this can be desirable.

Kylebrown9:
Created issue #236

PR #237, opened 2023-02-14
invalid span shadowing error message

### LINKED ISSUE #219: miette fails silently when span is wrong
OK I now understand why I just get an error message without the full beautiful report:
 
Screen Shot 2022-10-30 at 7 50 38 PM

it's because my span is wrong. It would have been nice if an error would have been thrown so that it's easier to understand why no report is displayed : o

Mehrbod2002:
Fixes #219 
Sorry for non-test PR as it should be announce and throw it , because of my leak familiar on crate .
Hope owner or collaborator add a test on it with help .

### DIFF
--- a/miette-derive/src/label.rs
+++ b/miette-derive/src/label.rs
@@ -114,10 +114,9 @@ impl Labels {
                 quote! {
                     miette::macro_helpers::OptionalWrapper::<#ty>::new().to_option(&self.#span)
                     .map(|#var| miette::LabeledSpan::new_with_span(
-                            std::option::Option::Some(format!(#fmt #args)),
-                            #var.clone(),
-                        )
-                    )
+                        std::option::Option::Some(format!(#fmt #args)),
+                        #var.clone(),
+                    ))
                 }
             } else {
                 quote! {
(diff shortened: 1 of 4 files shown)


Benjamin-L:
This fix would only apply for labels that are part of a `Diagnostic` this was implemented with the derive macro, which covers most, but not all, uses.

I think the best place to do this check is in the `SourceCode::read_span` impls. It looks like the existing impl [already returns `MietteError::OutOfBounds` in some situations](https://github.com/zkat/miette/blob/main/src/source_impls.rs#L95).

Mehrbod2002:
> This fix would only apply for labels that are part of a `Diagnostic` this was implemented with the derive macro, which covers most, but not all, uses.
> 
> I think the best place to do this check is in the `SourceCode::read_span` impls. It looks like the existing impl [already returns `MietteError::OutOfBounds` in some situations](https://github.com/zkat/miette/blob/main/src/source_impls.rs#L95).

Thanks for helpful command !

Mehrbod2002:
@Benjamin-L I can we can try it out

Issue #239, opened 2023-03-01
`Severity` should implement `Ord`


ErichDonGubler:
I realize that I'm both (a) a newcomer to this community and (b) haven't proposed this idea anywhere. I figured a PR was the most straightforward way to communicate the idea, but discovered that the preferred flow as an issue first, so I have both in hand already. 😅

---

I have a use case where accumulate the “highest” severity observed during analysis of some artifacts for which I report diagnostics with `miette`. Slicing some code that I was working with (not up on GH yet) into a reduced example that I'd _like_ to work:

```rust
struct AnalysisStats {
    highest_severity_seen: Option<Severity>,
    // Maybe some other state…
}

impl AnalysisStats {
    fn track<T>(&mut self, t: T) -> T
    where
        T: Debug + Diagnostic + Display,
    {
        self.highest_severity_seen = Some(
            self.highest_severity_seen
                .unwrap_or(Severity::Error)
                .max(t.severity()),
        );
        t
    }
}
```

The simplest work-around for this issue I've found so far is to make a hidden `enum` that mirrors `Severity` and derive `Ord` on it, viz. something like this:

```rust
fn track<T>(&mut self, t: T) -> T
where
    T: Debug + Diagnostic + Display,
{
    #[derive(Eq, Ord, PartialEq, PartialOrd)]
    enum SeverityOrd {
        Error,
        Warning,
        Advice,
    }

    impl From<Severity> for SeverityOrd {
        fn from(value: Severity) -> Self {
            match value {
                Severity::Error => SeverityOrd::Error,
                Severity::Warning => SeverityOrd::Warning,
                Severity::Advice => SeverityOrd::Advice,
            }
        }
    }

    impl From<SeverityOrd> for Severity {
        fn from(value: SeverityOrd) -> Self {
            match value {
                SeverityOrd::Error => Severity::Error,
                SeverityOrd::Warning => Severity::Warning,
                SeverityOrd::Advice => Severity::Advice,
            }
        }
    }

    let severity = t.severity().unwrap_or(Severity::Error);
    self.highest_severity_seen =
        Some(self.highest_severity_seen.clone().map_or(severity, |s| {
            SeverityOrd::from(s).max(severity.into()).into()
        }));
    t
}
```

The above is an unfortunate amount of code. 😅 I believe `Severity` should implement `Ord`, because:

1. The notion of the “height” of severity is fairly clear (e.g., higher is worse).
2. Accumulating severity like this is not uncommon in applications that I've worked with.

[...]

PR #240, opened 2023-03-01
feat(protocol): implement `Ord` for `Severity`

ErichDonGubler:
Implements #239, pending acceptance.


zkat (approved):
This seems perfectly reasonable to me! Thanks for the patch!

PR #244, opened 2023-03-26
refactor: nit perf improvements

YoniFeng:
While investigating a perf issue (https://github.com/Boshen/oxc/issues/166) when many miette errors were printed, before I understood the cause (miette is not the problem) I made these local changes.


zkat (approved):
Sure why not. Seems fine enough :)

PR #247, opened 2023-03-29
update to syn2

jtroo:


miette-derive/src/label.rs
```diff
@@ -88,7 +99,8 @@ impl Labels {
                             span: field.span(),
                         })
                     };
-                    let LabelAttr { label } = syn::parse2::<LabelAttr>(attr.tokens.clone())?;
+                    use quote::ToTokens;
+                    let LabelAttr { label } = syn::parse2::<LabelAttr>(attr.meta.to_token_stream())?;
```
jtroo, line 103: Notable here is that I didn't see a way to get a token stream that behaved exactly the same as in syn v1, but it seemed that using `meta` while skipping the first token was suitable, so that's what I went with.

zkat (approved):

zkat:
wow this is great! Thank you! I hope this doesn't cause any major breakage, but I guess we'll see 😅

zkat:
@jtroo looks like this just needs a quick rustfmt and we're good to go :)

jtroo:
`rustfmt` applied 👍

PR #250, opened 2023-04-08
feat: Add style `tags` field for graphical theme.

milesj:
Hey @zkat, great library.

We're in the process of migrating to miette from our own custom error handler and there's 1 feature we use heavily that miette does not support, and that's "style tags" within error messages. Here's an example of this:

```rust
#[error("Unable to unpack <path>{0}</path>, unsupported archive format <label>{1}</label>.")]
UnsupportedArchiveFormat(PathBuf, String),
```

It's basically HTML-like tags within the error message, that we replace with ANSI coloring/links/etc before rendering to the terminal. This is really nice as we can use these same tags in error messages, logs, and tracing, and all of them render in a similar fashion.

I originally approached this PR by supporting a "message formatting function", but because the graphical handler requires `Clone` + `Debug`, this wasn't easy to set and pass around. So instead I thought, why not just support the tags natively? As part of the theme perhaps? And that's what this PR does.

Here's an example of the terminal output when writing the test:
Screenshot 2023-04-08 at 4 39 12 PM

Another question I have:

- Should the json/narratable handlers remove all tags? Or preserve them? If yes, I think this one will require using regex.

### DIFF
--- a/tests/graphical.rs
+++ b/tests/graphical.rs
@@ -4,6 +4,7 @@ use miette::{
     Diagnostic, GraphicalReportHandler, GraphicalTheme, MietteError, NamedSource,
     NarratableReportHandler, Report, SourceSpan,
 };
+use owo_colors::Style;
 use thiserror::Error;
 
 fn fmt_report(diag: Report) -> String {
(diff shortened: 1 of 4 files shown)


tests/graphical.rs
```diff
@@ -1214,3 +1215,32 @@ fn zero_length_eol_span() {
...
+    let report: Report = err.into();
+    let mut out = String::new();
+
+    GraphicalReportHandler::new_themed(theme)
+        .with_width(80)
+        .render_report(&mut out, report.as_ref())
+        .unwrap();
+
+    println!("Error: {}", out);
+    let expected = "\u{1b}[31moops::my::bad\u{1b}[0m\n\n  \u{1b}[31m×\u{1b}[0m this \u{1b}[32mis\u{1b}[0m a \u{1b}[33mstring \u{1b}[34mwith\u{1b}[0m\u{1b}[0m many \u{1b}[32mstyle\u{1b}[0m tags!\n";
```
milesj, line 1243: wasnt sure if this was the best approach :P

zkat:
This is a cool idea! It definitely sounds nice to be able to have more formatting control over Graphical. That said, a change like this would be significantly breaking (for example, if a message renders actual xml/html as part of its message text).

I'm also not sure I feel comfortable messing with a field that sort of "belongs" to thiserror, if you know what I'm getting at. That particular error message is intended to work outside of any particular formatter.

In the end, I wonder if there's alternatives to this that might prove good enough, such as your end application deciding to apply these styles to its messages itself, since that's well within your purview and doesn't need to concern itself with what might conflict with a larger ecosystem. 

I can imagine something like this working pretty well, Willy the need to deal with markup:

```rust
#[error("thing: {}, other: {}", .0.fg::<Red>, .1.underline())]
MyError(String, String)
```

What do you think?

milesj:
Yeah that's fair, let me tinker around a bit with that approach.

milesj:
It's a bit more verbose, but works.  https://github.com/moonrepo/starbase/pull/17

We'll go with that approach, thanks! I always forget that extra syntax exists for thiserror.

Issue #236, opened 2023-02-12
Allow the inner source type of a `NamedSource` to be borrowed directly


esoterra:
I would like to be able to use `&NamedSource` as the input to parsing and lexing functions that use and borrow from the source already stored in the `NamedSource::inner()` instead of passing another copy.

Based on the discussion in #235, we could do this by making `NamedSource` generic over the inner source type which could simplify the API for a basic parse function to something like this:
```rust
pub fn parse(source: &NamedSource<String>) -> Result<Tokens<'_>> {
   let chars = source.inner().char_indices();
   ...
}
```

kallisti-dev:
Submitted a PR for this: #254 . See my comment there for notes about compatibility. I'll look into a backwards compatible version of this over the next few days. Feedback would be appreciated.

PR #254, opened 2023-04-23
Allow inner source type of a NamedSource to be borrowed directly

kallisti-dev:
Fixes: #236 

This is a simple direct implementation that just makes NamedSource generic over a SourceCode type parameter. It would be nice if we could give the type parameter a sensible default so that existing diagnostics don't break. I'll test various approaches to make this more backwards compatible over the next few days. 

Maybe keeping the source in a `Box`, allowing `S` to be `?Sized`, and then defaulting `S` to `dyn SourceCode + 'static` but I'm not sure how to change the type of `NamedSource::new` to allow for that, since function arguments must be `Sized`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ struct MyBad {
     // The Source that we're gonna be printing snippets out of.
     // This can be a String if you don't have or care about file names.
     #[source_code]
-    src: NamedSource,
+    src: NamedSource<String>,
     // Snippets and highlights can be included in the diagnostic!
     #[label("This bit here")]
     bad_bit: SourceSpan,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,7 +108,7 @@
 //!     // The Source that we're gonna be printing snippets out of.
 //!     // This can be a String if you don't have or care about file names.
 //!     #[source_code]
-//!     src: NamedSource,
+//!     src: NamedSource<String>,
 //!     // Snippets and highlights can be included in the diagnostic!
 //!     #[label("This bit here")]
 //!     bad_bit: SourceSpan,
--- a/src/named_source.rs
+++ b/src/named_source.rs
@@ -3,12 +3,12 @@ use crate::{MietteError, MietteSpanContents, SourceCode, SpanContents};
 /// Utility struct for when you have a regular [`SourceCode`] type that doesn't
 /// implement `name`. For example [`String`]. Or if you want to override the
 /// `name` returned by the `SourceCode`.
-pub struct NamedSource {
-    source: Box<dyn SourceCode + 'static>,
+pub struct NamedSource<S: SourceCode + 'static> {
+    source: S,
     name: String,
 }
 
-impl std::fmt::Debug for NamedSource {
+impl<S: SourceCode> std::fmt::Debug for NamedSource<S> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("NamedSource")
             .field("name", &self.name)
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -34,21 +34,24 @@ libraries and such might not want.
...
-  - [... in application code](#-in-application-code)
-  - [... in `main()`](#-in-main)
-  - [... diagnostic code URLs](#-diagnostic-code-urls)
-  - [... snippets](#-snippets)
-  - [... multiple related errors](#-multiple-related-errors)
-  - [... delayed source code](#-delayed-source-code)
-  - [... handler options](#-handler-options)
-- [Acknowledgements](#acknowledgements)
-- [License](#license)
+- [`miette`](#miette)
```
kallisti-dev, line 37: Whoops. My Markdown VSCode extension automatially updated this TOC. I'll disable that and recommit.

kallisti-dev:
Looks like you can make this backwards compatibility. But there are still a few issues to address with the `new` constructor.

You can change NamedSource to take a ?Sized parameter and then Box it, with the default being `dyn SourceCode + 'static`

```rs
pub struct NamedSource<S: SourceCode + ?Sized + 'static = dyn SourceCode + 'static> {
    source: Box<S>,
    name: String,
}
```

This allows `NamedSource` to refer to exactly the same type as before, so that this is no longer a breaking change.

However, in `NamedSource::new` you need to adjust the input parameter so that it converts properly. Here's what I came up with so far:

```rs
impl<S: SourceCode + ?Sized + 'static> NamedSource<S> {
    /// Create a new `NamedSource` using a regular [`SourceCode`] and giving
    /// its returned [`SpanContents`] a name.
    pub fn new(name: impl AsRef<str>, source: impl Into<Box<S>>) -> Self
    where
        S: Send + Sync,
    {
        Self {
            source: source.into(),
            name: name.as_ref().to_string(),
        }
    }
  }
```

This works fine for `NamedSource<String>` or `NamedSource<str>`, but for `NamedSource` it becomes too ambiguous and the compiler can't figure out which trait implementation to use.

```
error[E0283]: type annotations needed
   --> tests/graphical.rs:268:23
    |
268 |     .with_source_code(NamedSource::new("bad_file.rs", src));
    |                       ^^^^^^^^^^^^^^^^ cannot infer type for struct `Box<_>`
    |
    = note: multiple `impl`s satisfying `Box<_>: From<String>` found in the following crates: `alloc`, `miette`:
            - impl From<String> for Box<(dyn Diagnostic + 'static)>;
            - impl From<String> for Box<(dyn Diagnostic + Send + Sync + 'static)>;
            - impl From<String> for Box<(dyn std::error::Error + 'static)>;
            - impl From<String> for Box<(dyn std::error::Error + Send + Sync + 'static)>;
            - impl From<String> for Box<str>;
            - impl<T> From<T> for Box<T>;
    = note: required for `String` to implement `Into<Box<_>>`
note: required by a bound in `NamedSource::<S>::new`
   --> /Users/adam/Desktop/miette/src/named_source.rs:23:52
    |
23  |     pub fn new(name: impl AsRef<str>, source: impl Into<Box<S>>) -> Self
    |                                                    ^^^^^^^^^^^^ required by this bound in `NamedSource::<S>::new`

```

Is there a constraint that we could use instead of `Into<Box<S>>` that would work better here? And do we prefer this over the "simple but breaking` change to generics that's currently in the PR?

zkat:
I don't really see how we can work around that constraint, yeah, so this would necessarily need to be a breaking change :\

My recommendation continues to be that people write their own NamedSource implementation if they want something like this, considering that NamedSource is just so _trivial_

kallisti-dev:
Yeah, I don't have a particular opinion either way. I was just looking for simple issues to pick up and this one seemed like a quick fix. I do like how `NamedSource<String>` makes it a bit more clear what's underneath. I think it actually improves the docs a bit because it makes it a bit more clear in the examples what's going on. 

but I don't like that this is a breaking change for anything currently using `NamedSource`, and agree that writing a custom NamedSource would be very trivial.

MariaSolOs:
Ohhh I like this PR because I think it would help with #258 by making `NamedSource<T>` implement `Into<SourceSpan>` (which would be a trivial implementation delegating to the inner type).

MariaSolOs:
<s>Ohhh I like this PR because I think it can help with #254</s>.

nvm, it wouldn't. But doing something like this with `LabeledSpan` would. That is, changing it to:
```rust
pub struct LabeledSpan<S: Into<SourceSpan>> {
    label: Option<String>,
    span: S,
}
```

MariaSolOs:
To avoid the breaking changes, what if new generic types are added instead? The existing types can then just be specific variants of the new types.

zkat:
> To avoid the breaking changes, what if new generic types are added instead? The existing types can then just be specific variants of the new types.

Sorry, I have mom brain right now. What do you mean by this? Use a newtype or an alias? (I think using either of these might be breaking too??)

Issue #259, opened 2023-05-02
Add struct for dynamic creation of diagnostics


gavrilikhin-d:
It would be cool to have builtin way to create dynamic diagnostics in miette, instead of writing your own
```rs
#[derive(Debug, Error, PartialEq, Eq, Clone)]
pub struct CustomError {
    pub severity: Severity,
    pub message: String,
    pub code: Option<String>,
    pub help: Option<String>,
    pub labels: Option<Vec<LabeledSpan>>,
    pub url: Option<String>,
}

impl Display for CustomError {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.message)
    }
}

impl Diagnostic for CustomError {
    fn code<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
        self.code.as_ref().map(|s| Box::new(s) as Box<dyn Display>)
    }
    fn help<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
        self.help.as_ref().map(|s| Box::new(s) as Box<dyn Display>)
    }
    fn labels(&self) -> Option<Box<dyn Iterator<Item = miette::LabeledSpan> + '_>> {
        self.labels
            .as_ref()
            .map(|labels| Box::new(labels.iter().cloned()) as Box<dyn Iterator<Item = LabeledSpan>>)
    }
    fn severity(&self) -> Option<miette::Severity> {
        Some(self.severity)
    }
    fn url<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
        self.url.as_ref().map(|s| Box::new(s) as Box<dyn Display>)
    }
}
```

zkat:
I've actually talked with folks in the past about including something like this as a convenience! I'd totally take this as a PR :)

(you'll need to manually implement `std::error::Error`, ofc. `derive(Error)` is a `thiserror` thing and I don't want that dependency)

gavrilikhin-d:
What name should it have? My suggestions are:
- DynamicDiagnostic
- CustomDiagnostic
- MietteDiagnostic

Issue #201, opened 2022-09-01
`miette!` does not support capturing names in the current scope


gifnksm:
`miette!` does not support [`format_args_capture`] feature.

# Source

```rust
use miette::{miette, Result};

fn main() -> Result<()> {
    let x = 123;
    return Err(miette!("{x}"));
}
```

# Expected output

```
Error:
  × 123
```

# Actual output

```
Error:
  × {x}
```

[`format_args_capture`]: https://github.com/rust-lang/rust/issues/67984

PR #124, opened 2022-02-25
Docs overhaul

virtualritz:
* Fixed all intra-doc links and their formatting.
* Added clarification for `anyhow`/`eyre` users.
* Removed `MietteError::SetPrinterFailure` as it wasn't used any more and referenced a non-existing function (`set_printer()`).
* Moved the `README` text into `lib.rs` and generated `README.md` via `cargo readme` (with `README.tpl` template to have intra-doc links resolved correctly).
* Added `rustfmt.toml` to help with automatic line wrapping of comments via `cargo fmt`.
* Made the `ErrorHook` type public. It is referenced in the public docs and `set_hook()` (which is public) so I assumed this was intended.

PR #262, opened 2023-05-05
Add dynamic diagnostic

gavrilikhin-d:
Draft for dynamic diagnostics for design discussion. 
 
Implements #259

### DIFF
--- a/src/dynamic_diagnostic.rs
+++ b/src/dynamic_diagnostic.rs
@@ -1,0 +1,72 @@
+use std::{
+    error::Error,
+    fmt::{Debug, Display},
+};
+
+use crate::Diagnostic;
+
+/// Diagnostic that can be created at runtime.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DynamicDiagnostic {
+    /// Displayed diagnostic description
+    pub description: String,
+    /// Unique diagnostic code to look up more information
+    /// about this Diagnostic. Ideally also globally unique, and documented
+    /// in the toplevel crate's documentation for easy searching.
+    /// Rust path format (`foo::bar::baz`) is recommended, but more classic
+    /// codes like `E0123` will work just fine.
+    pub code: Option<String>,
+}
+
+impl Display for DynamicDiagnostic {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", &self.description)
+    }
+}
+
+impl Error for DynamicDiagnostic {}
+
+impl Diagnostic for DynamicDiagnostic {
+    fn code<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
+        self.code
+            .as_ref()
+            .map(Box::new)
+            .map(|c| c as Box<dyn Display>)
+    }
+}
+
+impl DynamicDiagnostic {
+    /// Create a new dynamic diagnostic with the given description.
+    ///
+    /// # Examples
+    /// ```
+    /// use miette::{Diagnostic, DynamicDiagnostic};
+    ///
+    /// let diag = DynamicDiagnostic::new("Oops, something went wrong!");
+    /// assert_eq!(diag.to_string(), "Oops, something went wrong!");
+    /// assert_eq!(diag.description, "Oops, something went wrong!");
+    /// ```
+    pub fn new(description: impl Into<String>) -> Self {
+        Self {
+            description: description.into(),
+            code: None,
+        }
+    }
+
+    /// Return new diagnostic with the given code.
+    ///
+    /// # Examples
+    /// ```
+    /// use miette::{Diagnostic, DynamicDiagnostic};
+    ///
+    /// let diag = DynamicDiagnostic::new("Oops, something went wrong!").with_code("foo::bar::baz");
+    /// assert_eq!(diag.description, "Oops, something went wrong!");
+    /// assert_eq!(diag.code, Some("foo::bar::baz".to_string()));
+    /// ```
+    pub fn with_code(self, code: impl Into<String>) -> Self {
+        Self {
+            code: Some(code.into()),
+            ..self
+        }
+    }
+}
--- a/src/eyreish/macros.rs
+++ b/src/eyreish/macros.rs
@@ -55,7 +55,7 @@
 /// fn divide(x: f64, y: f64) -> Result<f64> {
 ///     if y.abs() < 1e-3 {
 ///         bail!(
-///             severity = Severity::Warning;
+///             severity = Severity::Warning,
 ///             "dividing by value ({y}) close to 0"
 ///         );
 ///     }
@@ -64,16 +64,13 @@
 /// ```
 #[macro_export]
 macro_rules! bail {
-    ($err:expr $(,)?) => {
-        return $crate::private::Err($crate::miette!($err));
-    };
-    ($($key:ident = $value:expr),+; $($fmt:tt)+) => {
+    ($($key:ident = $value:expr,)* $fmt:literal $($arg:tt)*) => {
         return $crate::private::Err(
-            $crate::miette!($($key = $value),+; $($fmt)+)
+            $crate::miette!($($key = $value,)* $fmt $($arg)*)
         );
     };
-    ($($fmt:tt)+) => {
-        return $crate::private::Err($crate::miette!($($fmt)+));
+    ($err:expr $(,)?) => {
+        return $crate::private::Err($crate::miette!($err));
     };
 }
 
@@ -128,7 +125,7 @@ macro_rules! bail {
 /// fn divide(x: f64, y: f64) -> Result<f64> {
 ///     ensure!(
 ///         y.abs() >= 1e-3,
-///         severity = Severity::Warning;
+///         severity = Severity::Warning,
 ///         "dividing by value ({y}) close to 0"
 ///     );
 ///     Ok(x / y)
@@ -136,21 +133,16 @@ macro_rules! bail {
 /// ```
 #[macro_export]
 macro_rules! ensure {
-    ($cond:expr, $err:expr $(,)?) => {
-        if !$cond {
-            return $crate::private::Err($crate::miette!($err));
-        }
-    };
-    ($cond:expr, $($key:ident = $value:expr),+; $($fmt:tt)+) => {
+    ($cond:expr, $($key:ident = $value:expr,)* $fmt:literal $($arg:tt)*) => {
         if !$cond {
             return $crate::private::Err(
-                $crate::miette!($($key = $value),+; $($fmt)+)
+                $crate::miette!($($key = $value,)* $fmt $($arg)*)
             );
         }
     };
-    ($cond:expr, $fmt:expr, $($arg:tt)*) => {
+    ($cond:expr, $err:expr $(,)?) => {
         if !$cond {
-            return $crate::private::Err($crate::miette!($fmt, $($arg)*));
+            return $crate::private::Err($crate::miette!($err));
         }
     };
 }
@@ -167,7 +159,8 @@ macro_rules! ensure {
 /// let report = miette!("{x} + {} = {z}", y, z = x + y);
 /// assert_eq!(report.to_string().as_str(), "1 + 2 = 3");
 ///
-/// let report = miette!("{x} + {y} = {x + y}");
+/// let z = x + y;
+/// let report = miette!("{x} + {y} = {z}");
 /// assert_eq!(report.to_string().as_str(), "1 + 2 = 3");
 /// ```
 ///
@@ -182,8 +175,7 @@ macro_rules! ensure {
 ///     code = "expected::rparen",
 ///     help = "always close your parens",
 ///     labels = vec![LabeledSpan::at_offset(6, "here")],
-///     url = "https://example.com"; // <- semicolon separates fields from message
-///
+///     url = "https://example.com",
 ///     // Rest of the arguments are passed to `format!`
 ///     // to form diagnostic message
 ///     "expected closing ')'"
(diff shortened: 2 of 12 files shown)


gavrilikhin-d:
Do you agree with `DynamicDiagnostic` name?

gavrilikhin-d:
How generic should this class be? 

I tried using generic types with defaults:
```rs
struct DynamicDiagnostic<D = String, C = String, ...> 
where
    D: Debug + Display,
    C: Debug + Display,
    ...
{
   pub description: D,
   pub code: Option<C>,
   ...
}
```
But this was  ambiguous if using `None` values:
```rs
let diag: DynamicDiagnostic<&str, _, ...> = DynamicDiagnostic {
   description: "",
   code: None
};
```

src/dynamic_diagnostic.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct DynamicDiagnostic {
+    /// Displayed diagnostic description
+    pub description: String,
+    /// Unique diagnostic code to look up more information
+    /// about this Diagnostic. Ideally also globally unique, and documented
+    /// in the toplevel crate's documentation for easy searching.
+    /// Rust path format (`foo::bar::baz`) is recommended, but more classic
+    /// codes like `E0123` will work just fine.
+    pub code: Option<String>,
+}
```
zkat, line 19: Are we gonna have the other fields here too?

src/dynamic_diagnostic.rs
```diff
@@ -0,0 +1,72 @@
...
+    ///
+    /// # Examples
+    /// ```
+    /// use miette::{Diagnostic, DynamicDiagnostic};
+    ///
+    /// let diag = DynamicDiagnostic::new("Oops, something went wrong!");
+    /// assert_eq!(diag.to_string(), "Oops, something went wrong!");
+    /// assert_eq!(diag.description, "Oops, something went wrong!");
+    /// ```
+    pub fn new(description: impl Into<String>) -> Self {
```
zkat, line 49: I think the usual pattern is AsRef<str>, no?

src/lib.rs
```diff
@@ -249,7 +249,7 @@
 //! To construct your own simple adhoc error use the [miette!] macro:
 //! ```rust
 //! // my_app/lib/my_internal_file.rs
-//! use miette::{IntoDiagnostic, Result, WrapErr, miette};
+//! use miette::{miette, IntoDiagnostic, Result, WrapErr};
```
zkat, line 252: It would actually be really great to expand the `miette` macro as part of this PR so that you can do stuff like:

`miette!("something went wrong", code = foo::bar::bar)`

zkat (commented):
Looks good so far! I don't think this needs to be generic, tbh

src/lib.rs
gavrilikhin-d: Yes, agree.

src/dynamic_diagnostic.rs
gavrilikhin-d: Of course!

src/dynamic_diagnostic.rs
gavrilikhin-d: You mean `AsRef<str>`? But this will be less efficient if argument is already an owned `String`

src/miette_diagnostic.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl Diagnostic for MietteDiagnostic {
+    fn code<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
+        self.code
+            .as_ref()
+            .map(Box::new)
+            .map(|c| c as Box<dyn Display>)
+    }
+
+    fn severity(&self) -> Option<Severity> {
```
gavrilikhin-d, line 41: I would suggest to change `Diagnostic` to just return `Severity`, but this will be breaking change.
Using `Option<Severity>` is somewhat unreasonable, as every diagnostic haa some severity. And we can use `Severity::default()` instead of `None`

src/dynamic_diagnostic.rs
gavrilikhin-d: Almost everything from `Diagnostic` is added. I don't know wether `MietteDiagnostic` should (and can?) contain `diagnostic_source`, `related` and `source_code`.

gavrilikhin-d: These are dynamic over `Diagnostic` and `SourceCode` , that's why they would prevent `MietteDiagnostic` from implementing `PatialEq`, `Eq`, `Clone`, etc.

gavrilikhin-d: And this won't fit in my needs for example (I need to serialize/deserialize diagnostics with json and compare them for equality)

src/miette_diagnostic.rs
MariaSolOs: I think the current optional return type is useful in scenarios where you have an inner error proving must of the `miette` implementation.

For example:
```rust
#[derive(Debug, Error)]
#[error("{inner}")]
pub struct Outer {
  pub inner: Inner
}

impl miette::Diagnostic for Outer {
  fn severity(&self) -> Option<miette::Severity> {
    // I want all outer errors to be warnings by default
    // except if overridden by the inner error.
    self.kind.severity().or(Some(miette::Severity::Warning))
  }
}

#[derive(Debug, Diagnostic, Error)]
pub enum Inner {
  NotThatBad,
  
  #[diagnostic(severity("error"))]
  ReallyBad
}
```

gavrilikhin-d: Sounds reasonable

src/lib.rs
gavrilikhin-d: This conflicts with `format!` named arguments

gavrilikhin-d: My macro implementation:
```
    ($msg:literal, (code = $code:literal)? ) => {
        // Handle $:literal as a special case to make cargo-expanded code more
        // concise in the common case.
        crate::MietteDiagnostic::new($msg)
            $(.with_code($code))?
            .into()
    };
```

gavrilikhin-d: Oh, nevermind, it should be `$(code = $code:literal)? `. But this really could create confusion between named argument and parameter

src/eyreish/macros.rs
```diff
@@ -147,6 +144,23 @@ macro_rules! ensure {
 /// }
 /// ```
 ///
+/// With [`MietteDiagnostic`]-like arguments:
+/// ```
+/// use miette::{miette, LabeledSpan, Severity};
+///
+/// let source = "(2 + 2".to_string();
+/// let report = miette!(
```
gavrilikhin-d, line 152: Actually, I'd want to receive `MietteDiagnostic` here, instead of `Report`

gavrilikhin-d: Added `miette_diagnostic!`

src/lib.rs
gavrilikhin-d: Is it possible to add support for `miette!("{x} + {y}")` interpolation syntax without breaking others' code?

gavrilikhin-d: It was even requested in #201

zkat: Ahhhh that's a good point. I forgot about the format args thing!!

The way `tracing` resolves this is by treating key=value before the format strong as fields, and anything following the format string as format args. So it would look like this:

`miette!(code = foo::bar::baz, "error code from api: {code}", code = code)`

Where `code` is just a local var. 

How does that sound?

src/protocol.rs
```diff
@@ -218,6 +225,54 @@ impl LabeledSpan {
...
+    /// use miette::LabeledSpan;
+    ///
+    /// let source = "You have an eror here";
+    /// let label = LabeledSpan::underline(12..16);
+    /// assert_eq!(label, LabeledSpan::new(None, 12, 4))
+    /// ```
+    pub fn underline(span: impl Into<SourceSpan>) -> Self {
+        Self::new_with_span(None, span)
+    }
+
```
zkat, line 275: I like these a lot!!

src/protocol.rs
```diff
@@ -169,9 +169,16 @@ pub enum Severity {
     /// Warning. Please take note.
     Warning,
     /// Critical failure. The program cannot continue.
+    /// This is the default severity, if you don't specify another one
```
zkat, line 172:
```suggestion
    /// This is the default severity, if you don't specify another one.
```

src/miette_diagnostic.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// let diag = MietteDiagnostic::new("Typos in 'hello world'").with_labels(labels.clone());
+    /// assert_eq!(diag.description, "Typos in 'hello world'");
+    /// assert_eq!(diag.labels, Some(labels));
+    /// ```
+    pub fn with_labels(self, labels: Vec<LabeledSpan>) -> Self {
+        Self {
+            labels: Some(labels),
+            ..self
+        }
+    }
```
zkat, line 214: Can we have a version of these that appends labels instead of replacing them?

src/miette_diagnostic.rs
```diff
@@ -0,0 +1,215 @@
...
+    fmt::{Debug, Display},
+};
+
+use crate::{Diagnostic, LabeledSpan, Severity};
+
+/// Diagnostic that can be created at runtime.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MietteDiagnostic {
+    /// Displayed diagnostic description
+    pub description: String,
```
zkat, line 12: Can we rename this to `message`?

src/eyreish/macros.rs
```diff
@@ -162,7 +208,36 @@ macro_rules! miette {
...
+///     // Those fields are optional
+///     severity = Severity::Error,
+///     code = "expected::rparen",
+///     help = "always close your parens",
+///     labels = vec![LabeledSpan::at_offset(6, "here")],
+///     url = "https://example.com"
+/// );
+/// ```
+#[macro_export]
+macro_rules! miette_diagnostic {
```
zkat, line 237: This is kind of a mouthful. How would you feel about just calling it `diagnostic`?

zkat (commented):
This is looking really great so much! Thanks for taking this on! Just a couple more comments and I think we're good to merge?

src/lib.rs
gavrilikhin-d: Yep, if we want to support `format!("{x} + {y}")` thing, that moving fields before string is inevitable  anyway

src/miette_diagnostic.rs
gavrilikhin-d: Yes, but I can't figure out a proper name for it. It should also follow builder pattern for convenience, and `add_label` is imperative, more like a function that takes `&mut self`

gavrilikhin-d: Named it as `and_label`

src/eyreish/macros.rs
```diff
@@ -117,52 +146,113 @@ macro_rules! ensure {
...
         (&error).miette_kind().new(error)
     });
-    ($fmt:expr, $($arg:tt)*) => {
-        $crate::private::new_adhoc(format!($fmt, $($arg)*))
+    ($($key:ident = $value:expr),+; $($fmt:tt)+) => {
+        $crate::Report::from(
+            $crate::diagnostic!($($key = $value),+; $($fmt)+)
+        )
+    };
+    ($($fmt:tt)+) => {
```
gavrilikhin-d, line 214: First case with expression doesn't allow interpolation with single literal: 
```rs
let x = 1
let y = 2
miette!("{x} + {y}") // doesn't work as expected
miette!("{x} + {}", y) // works as expected
```

gavrilikhin-d: Do you have an idea, how we can work around this?

gavrilikhin-d: I feel like interpolation is more desirable than using expression in this macro, but removing it will be breaking change

src/miette_diagnostic.rs
zkat: Have you considered making these methods a sort of builder pattern, where they `mut self -> Self`? That's usually what I would expect this stuff to be, and it prevents unnecessary allocations, even if they're relatively small. You can always .clone(), after all.

src/eyreish/macros.rs
zkat: What if you used `$fmt:literal` and got rid of the semicolon thing? I think that would be good enough to disambiguate

gavrilikhin-d: Done. However, integer literals support (`miette!(0)`) is now lost. Can't find a way to differentiate them with string literals.

gavrilikhin-d:
is `cargo readme` from #124 not longer working?

zkat:
> is `cargo readme` from #124 not longer working?

@gavrilikhin-d `cargo make readme`

tests/test_chain.rs
```diff
@@ -1,7 +1,7 @@
 use miette::{miette, Report};
 
 fn error() -> Report {
-    miette!(0).wrap_err(1).wrap_err(2).wrap_err(3)
+    miette!("0").wrap_err(1).wrap_err(2).wrap_err(3)
```
zkat, line 4: Ooh is this a breaking change?

zkat (approved):
I think we're good now! I know this was a lot of back and forth but I think we got somewhere very nice. Thanks so much for your time and effort!!

zkat:
looks like you have a clippy thing too

tests/test_chain.rs
gavrilikhin-d: Well yes, but I doubt that someone uses miette like that

gavrilikhin-d:
Issues #259 and #201 will be resolved by this pr

tests/test_chain.rs
zkat: I'll just pretend I didn't see it ;)

gavrilikhin-d:
Why tests fail on windows? Looks like it can't find local arguments and assumes they are named arguments

zkat:
I think 1.56.0 is before format argument interpolation was stabilized, now that I think about it

gavrilikhin-d:
@zkat how can I fix it? See no way to conditionally compile code depending on rust version. 

My best guess is to introduce new default feature `format_args_capture`

gavrilikhin-d:
Now it will pass workflow (checked in my fork)

zkat:
Yay! It's in!!

PR #263, opened 2023-05-07
feat: Constify many functions

### LINKED ISSUE #261: Constify `SourceOffset` and `SourceSpan`
Since these are POD types containing simple integers, most of their code can be const. Specifically, all non-trait functions of `SourceSpan` and at least `SourceOffset::offset`. The other ones may use functions that are only const on nightly, and `SourceOffset::from_current_location` can most certainly not be const. However, these few would allow us to construct constant source spans and offsets, and extract their basic properties, which is my use case.

kleinesfilmroellchen:
This is primarily aimed at making `SourceSpan` and `SourceOffset` usable in const contexts.

Constifiable functions were found with the `clippy::missing_const_for_fn` lint, though it reported at least two false positives.

Fixes #261

### DIFF
--- a/src/eyreish/ptr.rs
+++ b/src/eyreish/ptr.rs
@@ -43,7 +43,7 @@ where
         Box::from_raw(self.ptr.as_ptr())
     }
 
-    pub(crate) fn by_ref<'a>(&self) -> Ref<'a, T> {
+    pub(crate) const fn by_ref<'a>(&self) -> Ref<'a, T> {
         Ref {
             ptr: self.ptr,
             lifetime: PhantomData,
@@ -91,7 +91,7 @@ where
         }
     }
 
-    pub(crate) fn from_raw(ptr: NonNull<T>) -> Self {
+    pub(crate) const fn from_raw(ptr: NonNull<T>) -> Self {
         Ref {
             ptr,
             lifetime: PhantomData,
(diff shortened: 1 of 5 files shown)


kleinesfilmroellchen:
Note: I couldn't test this with MSRV or any versions before 1.69 due to this error:
```
❯ cargo +1.58.0 nextest run      
    Updating crates.io index
error: failed to select a version for the requirement `aho-corasick = "^1.0.0"`
candidate versions found which didn't match: 0.7.20, 0.7.19, 0.7.18, ...
location searched: crates.io index
required by package `regex v1.8.1`
    ... which satisfies dependency `regex = "^1.5"` (locked to 1.8.1) of package `miette v5.8.0 (miette)`
error: command `'\\?\~\.rustup\toolchains\1.58.0-x86_64-pc-windows-msvc\bin\cargo.exe' test --no-run --message-format json-render-diagnostics` exited with code 101
```

zkat (approved):

zkat:
looks like this is relying on stuff that was unstable as of the current MSRV (1.56.0). Considering you can't seem to run this on earlier versions to test, maybe it's better to mark this as "breaking" and include it with miette 6.0 instead?

kleinesfilmroellchen:
> looks like this is relying on stuff that was unstable as of the current MSRV (1.56.0). Considering you can't seem to run this on earlier versions to test, maybe it's better to mark this as "breaking" and include it with miette 6.0 instead?

I didn't change my toolchain so these failures are strange. I will just throw away all the constification that cannot be done right now.

Issue #260, opened 2023-05-02
Use serde


gavrilikhin-d:
Could you make your `SourceOffset`, `SourceSpan`, `Severity`, `LabledSpan` and other types to implement `Serialize` and `Deserialize`?
Now you need to use builtin types instead of them to have serialization autoimplemented:
```rs
#[derive(Debug, Error, Diagnostic, Serialize, Deserialize, PartialEq, Eq, Clone)]
#[error("expected '{expected}'")]
pub struct Expected {
    pub expected: String,
    #[label("{expected}")]
    pub at: usize, // This is auto serialized/deserialized
}
```
So you can't just write:
```rs
#[derive(Debug, Error, Diagnostic, Serialize, Deserialize, PartialEq, Eq, Clone)]
#[error("expected '{expected}'")]
pub struct Expected {
    pub expected: String,
    #[label("{expected}")]
    pub at: SourceOffset, // Doesn't work. Need to implement manually
}
```

kleinesfilmroellchen:
Since serde is a large dependency, this should most definitely be an optional or opt-out dependency for anyone who doesn't need it.

zkat:
This seems reasonable, as long as it's hidden behind a non-default `serde` flag. I can't really do a lot of stuff myself right now, but I probably have time to review/merge a PR if someone contributes it.

gavrilikhin-d:
Ok, I will create a PR on these weekends

gavrilikhin-d:
This one looks related: #245

PR #264, opened 2023-05-13
Add `serde` support

gavrilikhin-d:
Need to add serde to `MietteDiagnostic` too, when #262 is merged

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -234,6 +234,7 @@ pub trait SourceCode: Send + Sync {
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct LabeledSpan {
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
     label: Option<String>,
     span: SourceSpan,
 }
@@ -337,7 +338,6 @@ fn test_serialize_labeled_span() {
     assert_eq!(
         json!(LabeledSpan::new(None, 0, 0)),
         json!({
-            "label": null,
             "span": { "offset": 0, "length": 0 }
         })
     );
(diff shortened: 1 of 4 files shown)


gavrilikhin-d:
Implements #260

src/protocol.rs
```diff
@@ -347,23 +419,22 @@ impl<'a> SpanContents<'a> for MietteSpanContents<'a> {
...
     offset: SourceOffset,
-    /// The total length of the span. Think of this as an offset from `start`.
-    length: SourceOffset,
+    /// The total length of the span
+    length: usize,
 }
 
 impl SourceSpan {
     /// Create a new [`SourceSpan`].
     pub fn new(start: SourceOffset, length: SourceOffset) -> Self {
```
gavrilikhin-d, line 434: It should be:
```suggestion
    pub fn new(start: SourceOffset, length: usize) -> Self {
```
But will be breaking change. 

`SourceOffset` may not represent length in no way. This looks more like a range (e.g `3..3`)

zkat: Can you make a pr for this once this lands? I'll mark it as breaking but that way we won't forget to do it. 

There's a good chunk of desirable breaking changes in the pipeline right now and I think I'm gonna find some time to roll up 6.0 at some point.

gavrilikhin-d: Ok

zkat:
Looks like you need to fix a conflict, but also I'm not sure about those other dependencies you bumped? Those seem to be what's causing the CI issues

zkat:
Also you may as well add serde support to MietteDiagnostic while you're at it

zkat:
Why do you need to update those dependencies?

gavrilikhin-d:
I do't need to :) Just noticed they are outdated and thought it may easy to fix it

PR #266, opened 2023-05-14
doc: document `#[diagnostic(transparent)]`

gavrilikhin-d:


src/lib.rs
```diff
@@ -198,6 +198,18 @@
 //!     #[error("Oops it blew up")]
 //!     #[diagnostic(code(my_lib::bad_code))]
 //!     BadThingHappened,
+//!
+//!     #[error(transparent)]
```
gavrilikhin-d, line 202: Shouldn't `#[diagnostic(transparent)]` imply `#[error(transparent)]`?

gavrilikhin-d:
`cargo make readme` doesn't work:
```
$ cargo make readme
error: no such command: `make`

        Did you mean `add`?

        View all installed commands with `cargo --list`
```

zkat:
You need to install cargo-make first

src/lib.rs
zkat: miette is blissfully unaware of thiserror. They are completely separate macros, and miette doesn't actually do anything related to `std::error::Error` itself.

milesj: @zkat If you're using darling (https://github.com/TedDriggs/darling) you can parse `#[error]` attributes and infer settings from it. This would help to avoid all the duplicate info in some cases.

gavrilikhin-d:
@zkat Hi! I've updated readme. Do I need to do something else to get this merged?

zkat (approved):

PR #270, opened 2023-05-29
documentation: updated example image (fixes #111)

### LINKED ISSUE #111: README example is discontinuous
https://github.com/zkat/miette/blob/2649fd27c47893dc3ba2445a9932600d1b3d3e63/README.md#L82-L162

The presented output image is one from nu:

miette error rendering of nu::parser::unsupported_operation

but the example is instead for rendering

miette error rendering of oops::my::bad

(Sorry, I can't do the pretty terminal window rendering)

TrapinchO:
I can get a larger image if necessary.


zkat:
I would prefer a larger image, and also one that shows off the error linkification

TrapinchO:
Increased image size as requested.

I am unsure what do you mean by the linkification. Do you want me to change the link text in the example (to show "link" or something similar instead) or to update `code_linking.png`?
Sadly my command line doesn't seem to be able to display links like that (I am on Windows), so I can't show it off.

zkat:
@TrapinchO the new Windows Terminal supports the feature:

image

TrapinchO:
I meant link tooltip, as seen on the current `code_linking` image.

I am afraid I don't understand what you want me to do. The error already includes a `url`. The other image is large enough and is about a different error.

zkat:
The problem is that it's displaying a URL at all. It should show the dashed underline instead.

TrapinchO:
This is the best I got to, no underline (`url("(link)")`). Cannot test whether it works due to windows terminal.
img

I would like to note that replacing `docsrs` with a string seems to display it whole and not just `(link)` and the documentation doesn't seem to show how to hide the link.

willbush:
Was thinking to make the same PR, but I see this is already made. I can make an image and share here or make another PR if you want?

Seems the `` in the readme should change too. And I noticed `lib.rs` has a copy of the readme. I tried replacing it with `#![doc = include_str!("../README.md")]`, but then tests started failing so I'm guessing that's why.

TrapinchO:
I have no problem with either. If you post it, I will update the pr, but feel free to open a fresh one.

willbush:
single-line-example

Here you go. The underline on `oops::my::bad (Files)` only shows up when I mouse over it, at least on Alacritty terminal in Linux. I had to zoom way in to get the resolution higher, mouse over the link, and take the screenshot with `grim`, and then crop the result in gimp.

zkat:
This latest one looks good to me, thanks :)

PR #273, opened 2023-06-16
fix(protocol): add StdError impl for Box<dyn Diagnostic + Send + Sync>

### LINKED ISSUE #272: `#[source]` does not work on `Box<dyn Error + Send + Sync>`
In order to get miette (with thiserror) to render the full error chain I need to be able to add the `#[source]` macro onto the "source" error.

In my project I'm using a boxed error that also needs to be `Send + Sync` like so:

```rust
#[derive(Diagnostic, Error)]
struct MyError {
  #[error("Any Error")]
  #[diagnostic(transparent)]
  AnyError(#[source] Box<dyn Diagnostic + Send + Sync>)
}
```

However that example fails with the following compilation error:

```
error[E0599]: the method `as_dyn_error` exists for reference `&Box<dyn Diagnostic + Send + Sync>`, but its trait bounds were not satisfied

20  |   pub trait Diagnostic: std::error::Error {
    |   ---------------------------------------
    |   |
    |   doesn't satisfy `dyn Diagnostic + Send + Sync: AsDynError<'_>`
    |   doesn't satisfy `dyn Diagnostic + Send + Sync: Sized`
```

I was able to get around the issue by using an `Arc` instead of a `Box` - but I don't need any `Arc` features so it would be great if it worked with `Box`.

bennetthardwick:
Adds `std::error::Error` impls for `Box<dyn Diagnostic + Send + Sync>` (and other variants).

Fixes #272.

### DIFF
--- a/tests/test_derive_source_chain.rs
+++ b/tests/test_derive_source_chain.rs
@@ -1,0 +1,69 @@
+use std::sync::Arc;
+
+use miette::{miette, Diagnostic, Report};
+use thiserror::Error;
+
+#[test]
+fn test_source() {
+    #[derive(Debug, Diagnostic, Error)]
+    #[error("Bar")]
+    struct Bar;
+
+    #[derive(Debug, Diagnostic, Error)]
+    #[error("Foo")]
+    struct Foo {
+        #[source]
+        bar: Bar,
+    }
+
+    let e = miette!(Foo { bar: Bar });
+    let mut chain = e.chain();
+
+    assert_eq!("Foo", chain.next().unwrap().to_string());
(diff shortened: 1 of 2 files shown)


tests/test_derive_source_chain.rs
```diff
@@ -0,0 +1,69 @@
+use std::sync::Arc;
+
+use miette::{miette, Diagnostic, Report};
```
zkat, line 3: looks like tests are failing because of this `Report` being unused.

bennetthardwick: Oops, fixed!

PR #275, opened 2023-06-24
docs: fix broken links

iceghost:
the links point to structs, whereas it should be traits


iceghost:
oh wait, i should fix it in `README.tpl`

zkat (approved):

PR #283, opened 2023-08-05
chore: update dependencies to their latest version

dnaka91:
Updated `textwrap` and `terminal_size` to their latest version, as those two crates had a breaking release.

Most changes seem to be simply fixes and breaking changes on API that are not used in miette.


zkat:
I've done this separately. Thank you.

PR #290, opened 2023-09-18
Improve ci and fix clippy

waywardmonkeys:


waywardmonkeys:
@zkat I agree! I can argue either side of it here, so I didn't know what to do. We can at least get most things fixed and I'll be happy to do what's needed in a follow up PR.

`rustix` is causing the problem and that's used by `is_terminal` ... so looking at that, `is_terminal` bumped MSRV in 0.4.8 from 1.48 to 1.63 where it is now (0.4.9).

We could modify the build for the MSRV to use `is_terminal` 0.4.7 perhaps?

zkat:
ugh. See, it feels incredibly rude to do an MSRV bump in a _patch release_.

I'm ok with locking our `is_terminal` version to `0.4.7` and just create an issue for removing the lock when we're ready to do a semver-major for miette.

Benjamin-L:
You might want to consider committing a lockfile that has version 0.4.7 of `is-terminal`, without bumping the version in `Cargo.toml`. My understanding is that a crate is considered to be compatible with a particular MSRV as long as it is possible to select dependency versions that are compatible with the MSRV and the constraints in `Cargo.toml`. Cargo usually won't find this set automatically. Hopefully some day we'll have a [MSRV-aware solver in cargo](https://github.com/rust-lang/cargo/issues/9930) and this whole problem will go away :)

zkat (approved):

waywardmonkeys:
@Benjamin-L Interesting ... maybe a possible follow up?

Benjamin-L:
I think pinning the dependency in `Cargo.toml` is a breaking change. For example, the following now fails to resolve:

```toml
[package]
name = "miette-290-test"
version = "0.1.0"
edition = "2021"

[dependencies]
is-terminal = "0.4.8"

[dependencies.miette]
git = "https://github.com/waywardmonkeys/miette.git"
branch = "improve-ci-and-fix-clippy"
features = [ "fancy" ]
```

The [semver docs][1] don't say anything about this situation, however the [cargo dependency docs][2] discourage upper bounds on library crates for this reason.

[1]: https://doc.rust-lang.org/cargo/reference/semver.html
[2]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-requirements

waywardmonkeys:
It is after midnight here, so tomorrow, I will look at fixing this in the CI scripts, instead of pinning the version like I did here.

waywardmonkeys:
@zkat @Benjamin-L I pushed a different way to resolve it. What do you think?

zkat (approved):

zkat:
I'm fine with this.

Issue #208, opened 2022-10-08
It is unclear how to display a line/column number on a NamedSource


brassy-endomorph:
In the output of the following code, I would expect the line/column numbers following the file name to be equal to the location of the error, but this isn't the case and from looking at the examples, I can't find a way to do this. I would be nice if this was documented somehow as the screenshots in the read me seem to have this error as well.

# Minimal example to reproduce

```bash
$ rustc --version
rustc 1.64.0 (a55dd71d5 2022-09-19)
$ cargo --version
cargo 1.64.0 (387270bc7 2022-09-16)
```

### `Cargo.toml`

```toml
[package]
name = "miette-test"
version = "0.1.0"
edition = "2021"
publish = false

[dependencies]
miette = { version = "5.3.0", features = ["fancy"] }
thiserror = "1.0.37"
```
### `main.rs`

```rust
use miette::{Diagnostic, SourceSpan, NamedSource, Report};

pub type Result<T> = ::std::result::Result<T, Error>;

#[derive(Diagnostic, thiserror::Error, Debug, PartialEq)]
#[error("Some error: {kind}")]
pub struct Error {
    pub kind: ErrorKind,
    #[label = "error"]
    pub source_span: SourceSpan,
}

#[derive(Diagnostic, thiserror::Error, Debug, PartialEq, Clone)]
pub enum ErrorKind {
    #[error("Something bad: {}", .0)]
    SomeKind(String),
}


fn main() -> miette::Result<()> {
    let program = "this is\na BAD program\n";
    Err(Error {
        kind: ErrorKind::SomeKind("oh no".to_owned()),
        source_span: (program.find("BAD").unwrap(), 3).into(),
    })  
    .map_err(Report::new)
    .map_err(|err| err.with_source_code(NamedSource::new("my_program.xyz", program)))
}
```

### Actual Terminal Output

```
Error: 
  × Some error: Something bad: oh no
   ╭─[my_program.xyz:1:1]
 1 │ this is
 2 │ a BAD program
   ·   ─┬─
   ·    ╰── error
   ╰────
```

### Expected Terminal Output

```
Error: 
  × Some error: Something bad: oh no
   ╭─[my_program.xyz:2:3]    <---- line 2 column 3
 1 │ this is
 2 │ a BAD program
   ·   ─┬─
   ·    ╰── error
   ╰────
```

CyriacBr:
This seems to be intended. See this https://github.com/zkat/miette/issues/91#issuecomment-950165566

[...]

Issue #91, opened 2021-10-23
Allow disabling of line/column display with an option


zen-xu:
```rust
use miette::{Diagnostic, SourceSpan};
use thiserror::Error;

#[derive(Error, Debug, Diagnostic)]
#[error("oops!")]
#[diagnostic(
    code(oops::my::bad),
    url(docsrs),
    help("try doing it better next time?")
)]
struct MyBad {
    // The Source that we're gonna be printing snippets out of.
    // This can be a String if you don't have or care about file names.
    #[source_code]
    src: NamedSource,
    // Snippets and highlights can be included in the diagnostic!
    #[label("This bit here")]
    bad_bit: SourceSpan,
}

/*
Now let's define a function!

Use this Result type (or its expanded version) as the return type
throughout your app (but NOT your libraries! Those should always return concrete
types!).
*/
use miette::{NamedSource, Result};
fn this_fails() -> Result<()> {
    // You can use plain strings as a `Source`, or anything that implements
    // the one-method `Source` trait.
    let src = "source\n  text\n    here".to_string();

    Err(MyBad {
        src: NamedSource::new("bad_file.rs", src),
        bad_bit: (9, 4).into(),
    })?;

    Ok(())
}

/*
Now to get everything printed nicely, just return a Result<()>
and you're all set!

Note: You can swap out the default reporter for a custom one using `miette::set_hook()`
*/
fn main() -> Result<()> {
    // kaboom~
    this_fails()?;

    Ok(())
}
```
![截屏2021-10-23 下午10 05 19](https://user-images.githubusercontent.com/38552291/138559867-c6ed03fd-e477-458e-a904-dc9146268a66.png)

Why it's `1:1` rather than `2:3`?

zkat:
not in the latest miette, no.

PR #291, opened 2023-09-19
Add support for primary label in specifying line/col information

willcrichton:
As discussed in #91 and #208, it is sometimes desirable for the line/column numbers in a diagnostic to reference a label as opposed to the start of the context. However, there has to be a way of identifying which label to reference.

This PR implements one possible mechanism: a `label(primary)` attribute which acts just like `label`, except it marks a single label as being "primary". For example, this program:

```rust
use miette::{Diagnostic, SourceSpan};
use thiserror::Error;

#[derive(Error, Debug, Diagnostic)]
#[error("Primary span error")]
struct PrimarySpanError {
    #[source_code]
    src: String,
    #[label(primary)] // note the new attribute parameter
    span: SourceSpan,
}

fn main() -> miette::Result<()> {
    Err(PrimarySpanError {
        src: String::from("Hello\nworld"),
        span: (7..10).into(),
    }
    .into())
}
```

Produces this output:

```
Error:   × Primary span error
   ╭─[2:2]
 1 │ Hello
 2 │ world
   ·  ───
   ╰────
```

Note that the line/column info is `2:2` rather than `1:1` which you would get with a normal `#[label]`.

The core technical changes are:
* Adding a `primary: bool` field to `LabeledSpan`.
* Adding a method `LabeledSpan::new_primary_with_span` that constructs a `LabeledSpan` with `primary: true`.
* Extending the derive macro to support `label(primary)`.
* Modifying the `GraphicalHandler` to look for a primary label and use it if possible.
* Added a test to `graphical.rs` to check for this behavior.

These changes were specifically made to be fully backwards compatible.

This is my first time in the miette codebase, so I may have missed a code path. Let me know if I should change anything.

### DIFF
--- a/miette-derive/src/label.rs
+++ b/miette-derive/src/label.rs
@@ -25,6 +25,7 @@ struct Label {
 
 struct LabelAttr {
     label: Option<Display>,
+    primary: bool,
 }
 
 impl Parse for LabelAttr {
@@ -41,10 +42,22 @@ impl Parse for LabelAttr {
             }
         });
         let la = input.lookahead1();
-        let label = if la.peek(syn::token::Paren) {
-            // #[label("{}", x)]
+        let (primary, label) = if la.peek(syn::token::Paren) {
+            // #[label(primary?, "{}", x)]
             let content;
             parenthesized!(content in input);
+
+            let primary = if content.peek(syn::Ident) {
+                let ident: syn::Ident = content.parse()?;
+                if ident != "primary" {
+                    return Err(syn::Error::new(input.span(), "Invalid argument to label() attribute. The argument must be a literal string or the keyword `primary`."));
+                }
+                let _ = content.parse::<Token![,]>();
+                true
+            } else {
+                false
+            };
+
             if content.peek(syn::LitStr) {
                 let fmt = content.parse()?;
                 let args = if content.is_empty() {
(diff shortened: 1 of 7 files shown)


zkat:
I think this kind of change is a long time coming and I really appreciate you taking the time to do this!

As far as specifics go: instead of having to add a new attribute, how would you feel about an API where, by default, the first attribute tagged with `#[label]` is treated as the primary one, and if you want a subsequent attribute to be the primary instead, you do `#[label(primary)]`? I think 99.9999% of the time, people don't really want the current "viewport" to be what shows up in that `[line:col]` display, and it's been an ongoing issue. I think a lot of people would be very happy if, without having to make any changes, that display just magically pointed to one of their spans, which will for the most part be the only one to begin with.

willcrichton:
I would be happy with that design as well. Two questions:
1. When you say "an API", do you mean a change to the default API everyone uses, or a distinct new part of the miette API surface?
2. If a user wants a format string in a primary label, should the expected syntax be `#[label(primary, "{}", arg)]`?

zkat:
1. Change `#[label]` itself so it starts marking things as "primary"
2. Yes. Basically in https://github.com/zkat/miette/pull/291/files#diff-a775e9b52316667cfcabc8d12a083cba8d18759534d48b7a5da6cd647bc6fe6dR394, to `.or_else(|| labels.first())` (or whatever)

willcrichton:
Ok I just updated the code and the tests. How does it look?

PR #300, opened 2023-10-05
Don't override provided source code

### LINKED ISSUE #299: `Report::with_source_code` overrides already provided source code
While working on compiler, I stumbled upon a panic from formatter. 
Turns out, using `with_source_code` overrides **EVERY** source codes, even already provided.

This is undesirable in multi-source environments. By intuition, `with_source_code` should only replace missing (not provided) source codes.

Here is a small example.
```rs
#[test]
fn miette_two_source_codes() -> miette::Result<()> {
    #[derive(thiserror::Error, miette::Diagnostic, Debug, Clone, PartialEq)]
    #[error("error")]
    struct E {
        #[label("here")]
        at: miette::SourceSpan,
        #[source_code]
        source_code: String,
    };

    let source1 = "abcdef";
    let source2 = "a";
    Err(miette::Report::from(E {
        at: (4..5).into(),
        source_code: source1.to_string(),
    })
    .with_source_code(source2.to_string()))
}
```

gavrilikhin-d:
Fixes #299

### DIFF
--- a/src/eyreish/wrapper.rs
+++ b/src/eyreish/wrapper.rs
@@ -163,7 +163,7 @@ impl<E: Diagnostic, C: SourceCode> Diagnostic for WithSourceCode<E, C> {
     }
 
     fn source_code(&self) -> Option<&dyn miette::SourceCode> {
-        Some(&self.source_code)
+        self.error.source_code().or(Some(&self.source_code))
     }
 
     fn related<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a dyn Diagnostic> + 'a>> {
@@ -197,7 +197,7 @@ impl<C: SourceCode> Diagnostic for WithSourceCode<Report, C> {
     }
 
     fn source_code(&self) -> Option<&dyn miette::SourceCode> {
-        Some(&self.source_code)
+        self.error.source_code().or(Some(&self.source_code))
     }
 
     fn related<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &'a dyn Diagnostic> + 'a>> {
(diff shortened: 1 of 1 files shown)


src/eyreish/wrapper.rs
```diff
@@ -232,3 +232,88 @@ impl<C> StdError for WithSourceCode<Report, C> {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use thiserror::Error;
+
+    use crate::{Diagnostic, LabeledSpan, Report, SourceCode, SourceSpan};
+
+    #[derive(Error, Debug)]
```
gavrilikhin-d, line 242: I couldn't use derive macro here. Looks like it always expands to `miette::`, instead of `crate::`. This may be fixed with `$crate::` in macro implementation, I think

src/eyreish/wrapper.rs
```diff
@@ -232,3 +232,88 @@ impl<C> StdError for WithSourceCode<Report, C> {
...
+                },
+                Inner {
+                    at: (1..2).into(),
+                    source_code: None,
+                },
+            ],
+        })
+        .with_source_code(outer_source.to_string());
+
+        let message = format!("{:?}", report);
```
gavrilikhin-d, line 315: Debug printing without `fancy` feature is broken

src/eyreish/wrapper.rs
zkat: This would be a good patch to have

zkat (approved):

anuejn:
Hey, this sadly actually breaks my use of miette :(.

I use kdl-rs and would like to be able to have the correct filename be printed in parser errors via a `NamedSource`. [Other users seem to have a similiar issues](https://github.com/kdl-org/kdl-rs/issues/134). I propose to add a method named `override_source_code` to miette that overrides existing code. Should I prepare a PR for that?

PR #301, opened 2023-10-05
Allow optional sources in derive

### LINKED ISSUE #217: SourceCode on Option<impl SourceCode>?
I would find it useful if there were such an implementation:

```rust
impl miette::SourceCode for Option<NamedSource> {
    fn read_span<'a>(
        &'a self,
        span: &miette::SourceSpan,
        context_lines_before: usize,
        context_lines_after: usize,
    ) -> std::result::Result<Box<dyn miette::SpanContents<'a> + 'a>, miette::MietteError> {
        self.expect("no source code was set in the error").read_span(span, context_lines_before, context_lines_after)     
    }
}
```

or perhaps a more general implementation:

```rust
impl<T> miette::SourceCode for Option<T> where T: miette::SourceCode {
    fn read_span<'a>(
        &'a self,
        span: &miette::SourceSpan,
        context_lines_before: usize,
        context_lines_after: usize,
    ) -> std::result::Result<Box<dyn miette::SpanContents<'a> + 'a>, miette::MietteError> {
        self.expect("no source code was set in the error").read_span(span, context_lines_before, context_lines_after)     
    }
}
```

this way I can have an optional source up until I set it:

```rust
#[source_code]
src: Option<NamedSource>
```

BTW I also noticed that I can't call this field `source` otherwise I get a weird error. Might be good to have a better error when I call the field `source` instead of `src`

gavrilikhin-d:
Resolves #217

### DIFF
--- a/miette-derive/src/source_code.rs
+++ b/miette-derive/src/source_code.rs
@@ -10,6 +10,7 @@ use crate::{
 
 pub struct SourceCode {
     source_code: syn::Member,
+    is_option: bool,
 }
 
 impl SourceCode {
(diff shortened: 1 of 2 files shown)


gavrilikhin-d:
@zkat Hi! Could you, please, share your thoughts about this PR? It's quite small, but very useful

zkat (approved):

Issue #303, opened 2023-10-15
`miette-derive` being required dependency


ManicMarrc:
Are there any reasons why `miette-derive` is a required dependency and not optional?
`miette-derive` is taking up a lot of my compile time but all I'm using is the `miette!` macro.

zkat:
I would take a patch that adds miette-derive as a default, but optional feature

zkat:
I don't know why github didn't close this issue. The related PR was merged. Thanks again.

PR #304, opened 2023-10-16
Make `miette-derive` be able to be turned off

ManicMarrc:
Implements #303

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -40,7 +40,8 @@ impl Diagnostic for MietteError {
     fn url<'a>(&'a self) -> Option<Box<dyn fmt::Display + 'a>> {
         let crate_version = env!("CARGO_PKG_VERSION");
         Some(Box::new(format!(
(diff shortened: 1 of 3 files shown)


zkat (approved):
thanks for putting this together!

src/error.rs
```diff
@@ -1,27 +1,46 @@
...
+            MietteError::OutOfBounds => Some(Box::new(
+                "Double-check your spans. Do you have an off-by-one error?",
+            )),
+        }
+    }
+
+    fn url<'a>(&'a self) -> Option<Box<dyn fmt::Display + 'a>> {
+        let crate_version = env!("CARGO_PKG_VERSION");
+        Some(Box::new(format!(
+            "https://docs.rs/miette/{crate_version}/miette/enum.MietteError.html"
```
zkat, line 43: btw, this should point to specific variants, like `https://docs.rs/miette/latest/miette/enum.MietteError.html#variant.IoError`

ManicMarrc: Ahh, okay

PR #308, opened 2023-10-21
Bump terminal-size to v0.3.0

notgull:


zkat:
is there a reason for this change? It looks like making it will require bumping the MSRV

notgull:
The new version of `terminal-size` uses `rustix` instead of `libc`, empowering it to use direct system calls on Linux and avoid linking to the C standard library. In addition, v1.63.0 is the version of Rust used by Debian Stable.

notgull:
@zkat I can bump the MSRV in this PR if need be.

notgull:
@zkat Any updates on this PR?

zkat:
Nothing new here--I'm going to sit on it until I think we're at a point where it's ok to do a semver-major bump on miette, because I only bump the MSRV with a major bump.

zkat:
It is time!

PR #312, opened 2023-11-02
Extend error text span to whole code points

MattX:
This fixes a panic when an error starts inside a Unicode code point. The range is extended to start (or end) at the beginning (or end) of the character inside which the byte offset is located.

Fixes zkat/miette#223. Unit tests provided in that issue by @Benjamin-L (lightly modified).

PR #314, opened 2023-11-04
Fix formatting bug when an empty span is not aligned to a char boundary

Benjamin-L:
This bug was originally identified in https://github.com/zkat/miette/pull/312#issuecomment-1791075099.

The bad output looked like this:

```
   ╭─[bad_file.rs:2:4]
 1 │ source
 2 │   👼🏼text
   ·   ─▲
   ·    ╰── this bit here
 3 │     here
   ╰────
```

Another possible option for the output (other than the one I implemented) would be

```
   ╭─[bad_file.rs:2:4]
 1 │ source
 2 │   👼🏼text
   ·    ▲
   ·    ╰── this bit here
 3 │     here
   ╰────
```

I prefer pointing the `▲` at the first column of the wide char rather than the second, but can see an argument that pointing at the second column is visually communicates that the span is unaligned. I think that's pretty niche, and would be surprised if any users are generating unaligned spans on purpose.

---

I also included a minor refactor in the PR, removing an unnecessary branch. The branch was always taken both before and after the bugfix. I can remove that bit if you want.

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -718,7 +718,11 @@ impl GraphicalReportHandler {
                 let byte_start = hl.offset();
                 let byte_end = hl.offset() + hl.len();
                 let start = self.visual_offset(line, byte_start, true).max(highest);
-                let end = self.visual_offset(line, byte_end, false).max(start + 1);
+                let end = if hl.len() == 0 {
+                    start + 1
+                } else {
+                    self.visual_offset(line, byte_end, false).max(start + 1)
+                };
 
                 let vbar_offset = (start + end) / 2;
                 let num_left = vbar_offset - start;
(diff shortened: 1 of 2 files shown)


src/handlers/graphical.rs
```diff
@@ -718,31 +718,33 @@ impl GraphicalReportHandler {
...
-                            chars.underline.to_string().repeat(num_right),
-                            width = start.saturating_sub(highest),
-                        )
-                        .style(hl.style)
-                        .to_string(),
-                    );
-                }
+                underlines.push_str(
+                    &format!(
+                        "{:width$}{}{}{}",
```
jdonszelmann, line 732: I just fixed a bug in #316 that also uses these `:width$` format specifiers. It seems that they're not such a good idea when working with ansi. I do see that you apply style after you do that here though, so it might work fine here actually. I just noticed it in this diff as well and maybe this should be checked everywhere :thinking:

Benjamin-L: I believe it's fine in this case, since the actual format arg is `""`. `format!("{:width$}", "", width = x)` should just expand to `x` space chars. I might be misunderstanding what the problem is, but I don't _think_ this is the same bug as the one you found in #316.

jdonszelmann: ohh yes that is correct, I do that too somewhere I think. You're right, this works :D

PR #316, opened 2023-11-08
rendering bug on small spans in large spans

### LINKED ISSUE #317: Rendering bug on small spans in large spans
While using miette, I found a small but significant rendering bug in rendering arrows. The picture below shows what it looks like:
image

Of course, I can't just give you just that image to go on, so I made a minimal example:

```rust
#[derive(Error, Debug, Diagnostic)]
#[error("oops!")]
#[diagnostic(severity(Error))]
struct MyBad {
    #[source_code]
    src: NamedSource,
    #[label("big")]
    big: SourceSpan,
    #[label("small")]
    small: SourceSpan,
}
let err = MyBad {
    src: NamedSource::new("issue", "\
if true {
    a
} else {
    b
}"),
    big: (0, 32).into(),
    small: (14, 1).into(),
};
let out = fmt_report(err.into());
println!("Error: {}", out);

```
This prints:
```rust
r#"  × oops!
   ╭─[issue:1:1]
 1 │ ╭─▶ if true {
 2 │ │╭▶     a
   · ││    ┬
   · ││    ╰── small
 3 │ │   } else {
 4 │ │       b
 5 │ ├─▶ }
   · ╰──── big
   ╰────
"#
```

while I expected it to print:

```rust
r#"  × oops!
   ╭─[issue:1:1]
 1 │ ╭─▶ if true {
 2 │ │       a
   · │       ┬
   · │       ╰── small
 3 │ │   } else {
 4 │ │       b
 5 │ ├─▶ }
   · ╰──── big
   ╰────
"#

```

As you can see, this looks a bit like a test, which is exactly what I wrote and contributed as a pull request so the bug is reproducible. See #316.

jdonszelmann:
See #317. This contributes a test for #317  and then fixes the bug :)

Closes #317 

The bad screenshot from #317 

image

now becomes this with the proposed changes:

image

Note that there were actually two subtle bugs. One was the double arrow in the gutter problem, which also occurred when rendering without ansi escape codes. However, this line
https://github.com/zkat/miette/blob/a8b4ae012aa0cf03b53a18f013c2b3f76c5040e7/src/handlers/graphical.rs#L611

Also did not work when the gutter contains ansi escape codes. It extends to the `width` variable, but when the string contains ansi codes it does not actually add any spaces since it thinks the escape codes have width as well. They of course do not, so the output doesn't look right. Unfortunately this is quite hard to make tests for.

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -410,7 +410,7 @@ impl GraphicalReportHandler {
         for line in &lines {
             let mut num_highlights = 0;
             for hl in &labels {
-                if !line.span_line_only(hl) && line.span_applies(hl) {
+                if !line.span_line_only(hl) && line.span_applies_gutter(hl) {
                     num_highlights += 1;
                 }
             }
@@ -529,7 +529,7 @@ impl GraphicalReportHandler {
         }
         let chars = &self.theme.characters;
         let mut gutter = String::new();
-        let applicable = highlights.iter().filter(|hl| line.span_applies(hl));
+        let applicable = highlights.iter().filter(|hl| line.span_applies_gutter(hl));
         let mut arrow = false;
         for (i, hl) in applicable.enumerate() {
             if line.span_starts(hl) {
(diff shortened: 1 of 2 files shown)


zkat:
Can you fix the conflicts here?

jdonszelmann:
@zkat there ya go! Thanks for having started such an awesome project :sparkling_heart:

PR #324, opened 2023-12-13
Clean up render snippets

LHolten:
Hey, here are some local changes I made that might be useful.

In the first commit I tried to make the `render_snippets` function more readable and fixed what I think was a bug.
(Looks like it would not merge more than two contexts if the last context didn't overlap the first)

In the second commit I restricted finding the primary label to those in the current context.
This makes it so that the header above each snippet points to code in the snippet.
(It is especially useful for `SourceCode` implementations that wrap multiple files)

Example showing different headers for different snippets:
image

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -6,7 +6,7 @@ use unicode_width::UnicodeWidthChar;
 use crate::diagnostic_chain::{DiagnosticChain, ErrorKind};
 use crate::handlers::theme::*;
 use crate::protocol::{Diagnostic, Severity};
-use crate::{LabeledSpan, MietteError, ReportHandler, SourceCode, SourceSpan, SpanContents};
+use crate::{LabeledSpan, ReportHandler, SourceCode, SourceSpan, SpanContents};
 
 /**
 A [`ReportHandler`] that displays a given [`Report`](crate::Report) in a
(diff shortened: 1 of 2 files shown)


src/handlers/graphical.rs
```diff
@@ -377,66 +377,63 @@ impl GraphicalReportHandler {
...
+                );
+                // Check that the two contexts can be combined
+                if let Ok(new_conts) =
+                    source.read_span(new_span.inner(), self.context_lines, self.context_lines)
+                {
+                    contexts.pop();
+                    contexts.push((
+                        // We'll throw this away later
+                        new_span,
+                        (new_conts.line(), new_conts.line_count()),
```
LHolten, line 425: We return the combined context lines here instead of only the context lines of only the first label to fix the bug mentioned in the PR description.
(`new_conts` contains the combined context lines and `left_conts` contains only context lines from the first label)

Benjamin-L (commented):
I'm not confident that the `x_conts.line(), x_conts.line_count()` -> `x_line, x_line_count` change is actually more readable personally, but don't feel strongly about it. It would probably also be good to add a test for the combined-context bug in `tests/graphical.rs`. The rest of this looks great to me.

LHolten:
> I'm not confident that the `x_conts.line(), x_conts.line_count()` -> `x_line, x_line_count` change is actually more readable personally, but don't feel strongly about it. It would probably also be good to add a test for the combined-context bug in `tests/graphical.rs`. The rest of this looks great to me.

I thought this change was necessary because I removed the vector of `SpanContents`. Turns out it was not necessary so I reverted the change and it is now even more readable!

LHolten:
I may have misunderstood the intended behaviour of `render_snippets`, it could be that it intentionally does not join snippets when the renderer would fail to render the combined context correctly.

LHolten:
Also added a test for non-adjacent contexts which shows that each context gets a header that points to the code in the context :D

Benjamin-L (approved):
New version looks really good to me :)

zkat (approved):

zkat:
Thanks for this!

PR #328, opened 2024-01-04
Add `wrap_lines: bool` option allowing wrapping be disabled entirely

zanieb:
Hi!

I want to be able to disable line wrapping entirely for logging runs and for users that would rather rely on their terminal's line wrapping. This pull request introduces a `wrap_lines` option which disables wrapping. Since `textwrap::fill` is used insert prefixes as well, this requires a little implementation to preserve that behavior.


zkat (approved):

zanieb:
Thanks!

PR #329, opened 2024-01-11
Improve robustness of color_format tests

Benjamin-L:
Previously, these tests would have spurious failures when `NO_COLOR` or `FORCE_COLOR` was set in the user's environment, since we weren't clearing one variable before testing a value for the other one. The previous version of the code also did not restore environment variable values on panic, which could cause spurious failures in other tests after one test fails.


zkat:
Thanks!

PR #330, opened 2024-01-27
fix(graphical): render cause chains for inner errors

TheLostLambda:
Hi @zkat!

The default `GraphicalReportHandler` disables the printing of cause chains for any inner errors (errors `related()` to a source diagnostic) when it disables nested footer printing. This results in lost cause chain information when printing with the default report handler.

Before this PR:
image

After this PR:
image

This PR simply re-enables that cause-chain printing by removing the line where it was disabled. It looks like that line was added as part of the initial feature implementation in #170, but the current behavior of not printing the cause chains of inner errors (I found) surprising and it led me to scour my codebase trying to find where the source error was being lost.

If, however, retaining the current default behavior is preferred, I can implement a couple of methods on `GraphicalReportHandler` like `.with_nested_cause_chains()` and `.without_nested_cause_chains()` — just let me know!

Also more than happy to write some tests after I know if you're happy with changing this default behavior or if you'd prefer a configuration option was added!

Hope you are well!
Brooks

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -317,9 +317,10 @@ impl GraphicalReportHandler {
                     ErrorKind::Diagnostic(diag) => {
                         let mut inner = String::new();
 
-                        // Don't print footer for inner errors
                         let mut inner_renderer = self.clone();
+                        // Don't print footer for inner errors
                         inner_renderer.footer = None;
+                        // Cause chains are already flattened, so don't double-print the nested error
                         inner_renderer.with_cause_chain = false;
                         inner_renderer.render_report(&mut inner, diag)?;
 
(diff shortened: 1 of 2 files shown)


TheLostLambda:
Whoops! Those tests didn't fail when I ran them locally, but I'll give then another look today! I'll get them passing!

TheLostLambda:
Turns out the line of code I deleted did have a reason to exist — go figure! I've added that back and turned back on the cause chains specifically for nested related errors. I've also added a test to capture the new behavior!

All tests are passing for me now (with `cargo test --all --features fancy` like is used in the CI)!

zkat (approved):

TheLostLambda:
Thanks @zkat !

PR #335, opened 2024-02-04
feat(fancy): Add option to change the link display texts

### LINKED ISSUE #331: Change `link` label to something else
I didn't find an easy way to do this, so I thought to open an issue to discuss if it's possible to implement.

I'd like to add documentation links to my errors, but the word `link` in the output doesn't sound correct. I was wondering if there is any way to provide a different label than `link`. In my case, I'd like it to say `documentation`.

This might be a minor nit, so feel free to disregard the issue if you don't consider it important. But if it's possible, it'd be great if someone could implement the option to change the default label for links.

calavera:
This option allows to globally change the default `(link)` display text with any other text provider by users.

I investigated if it was possible to change the display text per error, but I was not sold of any option I tried. I think this is a pretty simple addition that can help people like me that just want to change this text for all the errors.

Fixes #331

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -36,6 +36,7 @@ pub struct GraphicalReportHandler {
     pub(crate) word_separator: Option<textwrap::WordSeparator>,
     pub(crate) word_splitter: Option<textwrap::WordSplitter>,
     pub(crate) highlighter: MietteHighlighter,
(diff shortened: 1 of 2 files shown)


zkat:
They're minor issues--miette is outputting whitespace before the newline but autoformatting editors (like mine) remove trailing whitespace, so the strings end up not exactly matching. There's no actual problem. Merged :)

PR #338, opened 2024-02-07
feat(derive): enable more boxed types to be #[diagnostic_source]

TheLostLambda:
Hello again! Nice and small change this time :)

The `#[diagnostic_source]` attribute already worked for `Box<dyn Diagnostic>`, but not for the highly-related `Box<dyn Diagnostic + Send>` and `Box<dyn Diagnostic + Send + Sync>` types!

It looks like someone already solved this issue for `std::error:Error` with a `macro_rules!` implementation, so I've done the same here to let those new cases compile! Without those `Borrow<dyn Diagnostic>` impls, you end up with compiler errors coming from inside of the derive macro:

Screenshot from 2024-02-06 20-26-45

Let me know if you have any questions! I'll separately try to sort out those failing tests (unrelated to this change) in another PR :)

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -69,7 +69,7 @@ pub trait Diagnostic: std::error::Error {
     }
 }
 
-macro_rules! box_impls {
+macro_rules! box_error_impls {
     ($($box_type:ty),*) => {
         $(
             impl std::error::Error for $box_type {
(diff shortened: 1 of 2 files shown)


TheLostLambda:
With some luck, #339 should fix the tests failing for everyone — I don't think there was a behavior change, I think the reference strings just got mangled somehow in a commit from three days ago.

zkat (approved):
Awesome, thanks for this!

PR #340, opened 2024-02-08
feat(source): derive common traits for NamedSource, SourceSpan, and SourceOffset

TheLostLambda:
One-liner sanding off a small rough edge I came across: using `NamedSource<String>` (instead of simply `String`) severely limits which traits can be derived for the wrapping diagnostic. In my case, I lost the ability to clone my error!

image

I've brought things in line with https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits, leaving out `Copy`, `Display`, and `Default`, since those derives are either not possible or don't feel useful. The ones I've added here make sure people can `.clone()` errors, compare them, sort them, and use them in `HashSet`s / `HashMap`s.

I also added `Ord` derives to the `SourceSpan` and `SourceOffset` types for the same reason!

### DIFF
--- a/src/named_source.rs
+++ b/src/named_source.rs
@@ -3,6 +3,7 @@ use crate::{MietteError, MietteSpanContents, SourceCode, SpanContents};
 /// Utility struct for when you have a regular [`SourceCode`] type that doesn't
 /// implement `name`. For example [`String`]. Or if you want to override the
 /// `name` returned by the `SourceCode`.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct NamedSource<S: SourceCode + 'static> {
     source: S,
     name: String,
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -550,7 +550,7 @@ impl<'a> SpanContents<'a> for MietteSpanContents<'a> {
 }
 
 /// Span within a [`SourceCode`]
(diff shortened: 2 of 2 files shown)


src/named_source.rs
```diff
@@ -3,6 +3,7 @@ use crate::{MietteError, MietteSpanContents, SourceCode, SpanContents};
 /// Utility struct for when you have a regular [`SourceCode`] type that doesn't
 /// implement `name`. For example [`String`]. Or if you want to override the
 /// `name` returned by the `SourceCode`.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct NamedSource<S: SourceCode + 'static> {
```
zkat, line 7: I'm confused: doesn't this mean that `S` is now required to implement _all_ of those? I'm confused as to why they don't have to show up in the constraints. This is, of course, a silly Rust question.

TheLostLambda: Hopefully it's not too silly of a question — I had the same! It looks like the derive macros do the clever thing though and add a trait-bound on the generic — it means that if `S` implements `Clone`, then so will `NamedSource`, but if it doesn't, that's no big deal and `NamedSource` won't either.

`cargo-expand` shows: 
image

zkat: that's incredible!

zkat (approved):

PR #341, opened 2024-02-16
Allow collections as type for the label attribute (fixes #315)

### LINKED ISSUE #315: Is there a way to pull labels from a collection using proc macros?
Good day!

I have an enum with `#[derive(Diagnostic)]`.  I am trying to add an error type, which can have several spans sourced from a collection:

```Rust
pub enum ShellError {
    ...
    #[error("{0}")]
    #[diagnostic()]
    GenericError(
        String,
        #[label("{1}")]
        Vec<SourceSpan>,  // a vector with several spans
    ),
```

Is there a way to achieve this with proc macros?

Nahor:


zkat (approved):
You know what, this is really cool, so I'm not gonna pick on the syntax or whatever choices. It works and it's helpful/practical. Thanks so much!

Nahor:
You merged only one of the two commit. Any issue with the other (to allow LabeledSpan in the collections, mainly to be able to set the text for each individual label in a collection, since one can't use macro attributes in that case)?

zkat:
Yes, I squashed them. You're all good :)

PR #344, opened 2024-02-17
Simplify OptionalWrapper

Nahor:
1. That commit is based on #343 to avoid conflicts. But Github doesn't supports stacked PR across forks, so for now, it includes #343 as well. Just look at the last commit for the actual change.

2. I'm new to Rust and there are sometimes subtilties that escape me. So maybe it's not quite identical code. All the unit tests pass at least.

3. I don't know how you feel about minor changes like that. I would totally understand if you think it's a pointless/unnecessary one.


zkat:
On second thought, I'll keep this around as a breaking change for next time we do a semver-major.

PR #347, opened 2024-02-21
fix(invalid span): skip the snippet when read_span fails

Nahor:
Fixes: https://github.com/zkat/miette/issues/219

When a snippet couldn't be read (typically because the span didn't fit within the source code), it and the rest of the diagnostic were silently dropped, which was confusing to the developer.
Now, in place of the snippet, print an error message with the name of the failed label and the error it triggered, then proceed with the rest of the diagnostic (footer, related, ...)

This is what it looks like:
image

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -436,9 +436,25 @@ impl GraphicalReportHandler {
 
         let mut contexts = Vec::with_capacity(labels.len());
         for right in labels.iter().cloned() {
-            let right_conts = source
-                .read_span(right.inner(), self.context_lines, self.context_lines)
-                .map_err(|_| fmt::Error)?;
+            let right_conts =
+                match source.read_span(right.inner(), self.context_lines, self.context_lines) {
+                    Ok(cont) => cont,
+                    Err(err) => {
+                        writeln!(
+                            f,
+                            "  [{} `{}` (offset: {}, length: {}): {:?}]",
+                            "Failed to read contents for label".style(self.theme.styles.error),
+                            right
+                                .label()
+                                .unwrap_or("<none>")
+                                .style(self.theme.styles.link),
+                            right.offset().style(self.theme.styles.link),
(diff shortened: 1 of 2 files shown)


src/handlers/graphical.rs
```diff
@@ -436,9 +436,23 @@ impl GraphicalReportHandler {
...
-                .read_span(right.inner(), self.context_lines, self.context_lines)
-                .map_err(|_| fmt::Error)?;
+            let right_conts =
+                match source.read_span(right.inner(), self.context_lines, self.context_lines) {
+                    Ok(cont) => cont,
+                    Err(err) => {
+                        writeln!(
+                            f,
+                            "  [{} `{}`: {:?}]",
+                            "no snippet, error in label".style(self.theme.styles.error),
```
zkat, line 446:
```suggestion
                            "Failed to read contents for label".style(self.theme.styles.error),
```

src/handlers/graphical.rs
```diff
@@ -436,9 +436,23 @@ impl GraphicalReportHandler {
...
-            let right_conts = source
-                .read_span(right.inner(), self.context_lines, self.context_lines)
-                .map_err(|_| fmt::Error)?;
+            let right_conts =
+                match source.read_span(right.inner(), self.context_lines, self.context_lines) {
+                    Ok(cont) => cont,
+                    Err(err) => {
+                        writeln!(
+                            f,
+                            "  [{} `{}`: {:?}]",
```
zkat, line 445: We should print the source name, at least. Possibly the span, as well.

Nahor:
The CI failure should be addressed with #348

src/handlers/graphical.rs
Nahor: There is no name at this point. We only have a `SourceCode` trait object, which will give the name but only if the read_span succeeds.

zkat (approved):

Issue #346, opened 2024-02-21
Compile to wasm32-unknown-unknown / Compile without `terminal_size`


Boshen:
I'm currently stuck upgrading to v7 for wasm builds.

### Background

After the `terminal_size` upgrade from https://github.com/zkat/miette/pull/308, it brought in an extra dependency `rustix`, which in turn brought in `errrno` which doesn't support compile to `wasm32-unknown-unknown`.

```
error[E0583]: file not found for module `sys`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/errno-0.3.8/src/lib.rs:26:1
   |
26 | mod sys;
   | ^^^^^^^^
   |
   = help: to create the module `sys`, create file "/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/errno-0.3.8/src/sys.rs" or "/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/errno-0.3.8/src/sys/mod.rs"
   = note: if there is a `mod sys` elsewhere in the crate already, import it with `use crate::...` instead
```

After looking at both crates `terminal_size` and `errno`, both crates claims they don't support wasm.

### Possible Solution

Given that `terminal_size` is only called when terminal width is not provided, I think it's reasonable to put it behind a feature flag.

https://github.com/zkat/miette/blob/75fea0935e495d0215518c80d32dd820910982e3/src/handler.rs#L374-L381

zkat:
I'm ok putting it behind a feature flag but make it default as part of the fancy feature. Perhaps we can have a `fancy-wasm` that's specifically wasm-compatible?

zkat:
(that is, I'll take a patch to make these changes, and I can do a release after that for you)

Boshen:
I drafted a PR in https://github.com/zkat/miette/pull/349

I'm facing a problem with feature flag combinations. Currently all the "fancy" features are behind the "fancy-no-backtrace" flag:

https://github.com/zkat/miette/blob/6ea86a2248854acf88df345814b6c97d31b8b4d9/src/lib.rs#L776-L780

I can't figure out the best way to enable "fancy" for both `fancy-no-backtrace` and `fancy-no-syscall`

PR #349, opened 2024-02-22
feat: add feature "fancy-no-syscall" for wasm targets

Boshen:
* closes https://github.com/zkat/miette/issues/346
* closes https://github.com/zkat/miette/pull/326

I added 

* `fancy-base` feature to enable exports for the `fancy` feature
* `fancy-no-syscall` feature to disable all syscalls
* a `wasm` ci check to ensure future changes.

`target_family = "wasm"` will not work in this case because `terminal_size` will not compile under `wasm32-unknown-unknown`, the dependency has to be excluded.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,19 @@ jobs:
         if: matrix.rust == '1.70.0'
         run: cargo test --all --verbose --features ${{matrix.features}} no-format-args-capture
 
+  wasm:
+    name: Check Wasm build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


zkat (approved):

Boshen:
Thank you! I'll test this in my project tomorrow, let's make a release after I've confirmed the `fancy` features works on all systems.

Boshen:
I have verified both features `fancy-no-syscall` and `fancy-no-backtrace` are working as intended.

Boshen:
@zkat A new release would be greatly appreciated 🤝

zkat:
@Boshen done. Should be out in 7.2.0 now.

PR #358, opened 2024-03-27
fix(graphical): fix nested error wrapping

### LINKED ISSUE #333: Potentially broken word wrapping on `main`
Just a quick report, and I might have time to dig a bit deeper later on, but I've noticed a few times that I'll get some strange wrapping where the `GraphicalReportHandler` will move from the first to the second line, print exactly one word, then move to a third line.

I'm not particularly sure why yet, but it looks something like this: 
image
Here, `the` is the lonely word that gets banished to that middle line, though I think I've seen this with `a` as well, so it seems like one-char words are enough to trigger this issue. Resizing the terminal a bit causes the problem to disappear or reappear, so perhaps there is some sort of off-by-one, or several levels of wrapping that don't communicate indentation properly?

Hard to say, but I've left it here in case anyone else has more time to look into it than I do!

TheLostLambda:
Fixes #333

Here is an error now nicely wrapping (with the 2 char padding on either side) of a 50-char wide terminal:
image

### DIFF
--- a/src/eyreish/wrapper.rs
+++ b/src/eyreish/wrapper.rs
@@ -7,33 +7,7 @@ use crate::{Diagnostic, LabeledSpan, Report, SourceCode};
 use crate as miette;
 
 #[repr(transparent)]
-pub(crate) struct DisplayError<M>(pub(crate) M);
-
-#[repr(transparent)]
 pub(crate) struct MessageError<M>(pub(crate) M);
-
-pub(crate) struct NoneError;
-
-impl<M> Debug for DisplayError<M>
-where
-    M: Display,
-{
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        Display::fmt(&self.0, f)
-    }
-}
-
-impl<M> Display for DisplayError<M>
-where
-    M: Display,
-{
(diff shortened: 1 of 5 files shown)


TheLostLambda:
There is a second error here too that's come up! The prefixes for `#[related]` errors are messing up the line-width calculations:
image

The easiest solution seems to be changing the header to always add a newline after one of these tags — something that I think looks nicer anyways!
image

I'll also include that change in this PR!

TheLostLambda:
I've also fixed the spacing around / between related errors, and I'm trying to fix up some unrelated CI breakage (the new nightly compiler seems to be much better at / to much more aggressively detect dead code!)

TheLostLambda:
Oof, lots of little things making that slightly tricky, but ready for review now! I tried to keep some nice and meaningful commits if you're the type of person to review those diffs one-at-a-time!

TheLostLambda:
Finally, if you're looking for more example outputs and how they have changed / been restyled, here are plenty more examples: https://github.com/TheLostLambda/smithereens/commit/23a4093afc17c48eb7cda4f71faab44828719245

zkat (approved):

zkat:
Thanks for taking care of this! this looks a lot better :)

PR #363, opened 2024-04-05
Fix image in README.md

JakkuSakura:


zkat (approved):

zkat:
You need to make sure to make this change in lib.rs

PR #368, opened 2024-04-24
Turns out, `Report` *really* doesn't want to be an `Error` or `Diagnostic`...

TheLostLambda:
Failing some context tests, I'm pretty sure because `self.wrap_err(msg)` creates a `context_chain_downcast` table, whilst the `Report::from_msg(msg, self)` that it's impl clashed with only creates at `context_downcast`.

I messed around and got Rust to segfault, and discovered I was very much in over my head...

I do think it's possible to `impl Diagnostic for Report`, but there's a lot of internal restructuring that would need to happen, and *in the best case*, we lose the `Into<Report>` impl.

As with 90%+ of my issues with Rust nowadays, the stabilization of specialization would solve this (I believe), but that could be ages and ages away still...

Ultimately, this is here if anyone is feeling clever / brave, and it's work thinking about if we're okay losing those `From` and `Into` impls for `Report`, because I don't think there is a clever way around that...

### DIFF
--- a/.helix/languages.toml
+++ b/.helix/languages.toml
@@ -1,0 +1,2 @@
+[language-server.rust-analyzer]
+config = { cargo = { features = "all" } }
--- a/src/eyreish/context.rs
+++ b/src/eyreish/context.rs
@@ -1,4 +1,4 @@
-use super::error::{ContextError, ErrorImpl};
+use super::error::ContextError;
 use super::{Report, WrapErr};
 use core::fmt::{self, Debug, Display, Write};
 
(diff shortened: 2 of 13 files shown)


TheLostLambda:
If we decide this is better than have `From` and `Into`, I'll revisit with a couple of new ideas for how to solve the underlying issues, but for now I'll leave it be, since I'm not actually sure losing those `From` and `Into`s is worth it...

TheLostLambda:
Very happy to retire this if #371 is merged!

TheLostLambda:
Closing this in favour of the solution in #371 (and so that my only PRs still open are ones I believe should be merged!)

PR #372, opened 2024-04-29
fix(docs): `alt` attribut for `single-line-example.png`

skarab42:
The white line breaks in the `alt` attribute prevented the `single-line-example.png` image from being displayed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,11 +163,13 @@ Narratable printout:
 \
 diagnostic error code: oops::my::bad (link)
 Error: oops!
+
 \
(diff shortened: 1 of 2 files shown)


zkat:
you'll want to run `cargo make readme` to sync this up with `readme.md`

skarab42:
> you'll want to run `cargo make readme` to sync this up with `readme.md`

thanks for the tip, I'm new to Rust

zkat:
you're fine. Not everyone actually does things this way :)

zkat (approved):

PR #381, opened 2024-06-11
fix: setting NO_COLOR should not print ansi codes for non-terminals

### LINKED ISSUE #380: NO_COLOR is printing ansi in non-terminals
Background: https://github.com/oxc-project/oxc/issues/3539

> Currently, if you use oxc_diagnostics with GraphicalReportHandler, and you print to a non-terminal, the output will contain color escape codes, even with NO_COLOR=1. This is because if stdout or stderr are not a terminal, GraphicalTheme::default() will return GraphicalTheme::ascii() regardless of the NO_COLOR value, and ascii() does have colors. This means that when printing to a non-terminal, there is no way to turn off the color escapes.

> When printing to a non-terminal, you usually don't want color output, so presumably the default GraphicalTheme on non-terminals should be GraphicalTheme::none(). But in some cases it might be useful to enable color, even if not by default. Maybe FORCE_COLOR could be used in that case.

https://github.com/zkat/miette/blob/b82cc81b8ea32a1cf1b4598ed5832bc8e3b0e161/src/handlers/theme.rs#L72-L75

When it's non-terminal, `Self::ascii()` 

https://github.com/zkat/miette/blob/b82cc81b8ea32a1cf1b4598ed5832bc8e3b0e161/src/handlers/theme.rs#L25-L31

still uses ansi colors.

If the intention of `NO_COLOR` is to always not print ansi color codes, I can PR a change to `GraphicalTheme::none()`.

Boshen:
closes #380

### DIFF
--- a/src/handlers/theme.rs
+++ b/src/handlers/theme.rs
@@ -71,7 +71,7 @@ impl Default for GraphicalTheme {
     fn default() -> Self {
         match std::env::var("NO_COLOR") {
             _ if !std::io::stdout().is_terminal() || !std::io::stderr().is_terminal() => {
-                Self::ascii()
+                Self::none()
             }
             Ok(string) if string != "0" => Self::unicode_nocolor(),
             _ => Self::unicode(),


Boshen:
Rust version bumped from 1.77.2 to 1.78.0 in CI, do you want me to make a fixing PR?

zkat (approved):

zkat:
The CI failures don't have anything to do with your changes, so I'm just gonna merge this. Thanks a lot!

PR #382, opened 2024-06-12
fix: fix Rust v1.78.0 clippy warnings

Boshen:


src/eyreish/error.rs
```diff
@@ -410,7 +410,7 @@ impl Report {
     }
 
     /// Provide source code for this error
-    pub fn with_source_code(self, source_code: impl SourceCode + Send + Sync + 'static) -> Report {
+    pub fn with_source_code(self, source_code: impl SourceCode + 'static) -> Report {
```
Boshen, line 413: warning: this bound is already specified as the supertrait of `SourceCode`

Boshen:
🤔 Rust v1.78.0 also surfaced 2 unspecified features

```
error: unexpected `cfg` condition name: `doc_cfg`
 --> src/eyreish/mod.rs:1:13
  |
1 | #![cfg_attr(doc_cfg, feature(doc_cfg))]
  |             ^^^^^^^
  |
```

```
error: unexpected `cfg` condition name: `track_caller`
  --> src/eyreish/mod.rs:87:12
   |
87 | #[cfg_attr(track_caller, track_caller)]
   |            ^^^^^^^^^^^^
   |
```

Benjamin-L:
I think the `#[cfg_attr(track_caller, ...)]` usage was inherited from eyre. They attempt to probe stable `track_caller` support [in the build script][eyre-buildrs], and then set `cfg=track_caller` if it's available. We aren't doing this, but are [using][1] `track_caller` unconditionally already, and `track_caller` is available in our MSRV. I looked through all the uses of `#[cfg(track_caller)]` in miette, and believe we can just enable them all by default without a breaking change. I'd be in favor of doing that, although we should test it carefully.

`doc_cfg` is in a similar situation, where the code was inherited from eyre, and they set the cfg [in the cargo manifest][eyre-manifest]. We don't seem to be using `#[cfg(doc)]` anywhere, and this feature is also stable in our MSRV. We should just remove the `:#![cfg_attr(doc_cfg, feature(doc_cfg))]` line.

[eyre-buildrs]: https://github.com/eyre-rs/eyre/blob/907daa2cc667cb715874b2a743dec7592fc57484/eyre/build.rs#L55
[eyre-manifest]: https://github.com/eyre-rs/eyre/blob/907daa2cc667cb715874b2a743dec7592fc57484/eyre/Cargo.toml#L38
[1]: https://github.com/zkat/miette/blob/813232ba7957ae09e4fb9d9416d821f4fd9da66d/src/protocol.rs#L706

Boshen:
I've added `#![allow(unexpected_cfgs)]` so we can unblock other PRs.

I'll create an issue for removing these unexpected cfgs.

zkat (approved):

Boshen:
Ahh sorry I totally forgot about this PR, MIRI is failing due to 

```
error: unexpected `cfg` condition name: `track_caller`
  --> tests/test_fmt.rs:83:12
   |
83 | #[cfg_attr(track_caller, ignore)]
   |            ^^^^^^^^^^^^
```

You may `#[allow()]` it as well.

PR #385, opened 2024-06-20
feat(SourceSpan): add impl From<InclusiveRange>

Benjamin-L:
This can be used to avoid awkward `start..(end + 1)` constructions, which will trigger the `clippy::range_plus_one` lint.

I added a single impl for `InclusiveRange` rather than a blanket `impl<T: RangeBounds> From<T> for SourceSpan` for two reasons. The first is that the blanket impl would be a breaking change (because dependent crates may currently define a type `T: RangeBounds` and also have `impl From<T> for SourceSpan`). The second is that this would allow one-sided ranges (`..x`, `x..`). In order to support bounded-below ranges, we would need to change `SourceSpan` so that `length` may depend on the the specific `SourceCode` instance. That seems like an unlikely use case.

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -614,6 +614,28 @@ impl From<std::ops::Range<ByteOffset>> for SourceSpan {
     }
 }
 
+impl From<std::ops::RangeInclusive<ByteOffset>> for SourceSpan {
+    /// # Panics
+    ///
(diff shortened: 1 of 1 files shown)


Benjamin-L:
CI should be fixed by https://github.com/zkat/miette/pull/382

zkat (approved):

PR #387, opened 2024-06-26
perf(handlers): optimize string-buffer reallocations

lucab:
This improves `get_lines()` logic by using a string-buffer with a capacity hint. That avoids growing the buffer from zero each time, saving a bunch of reallocations.

This is a forwarded patch from https://github.com/oxc-project/oxc/pull/3897, see there for further explanations and datapoints.


zkat (approved):
Oh hey cool, thanks! I generally didn't spend a lot of time worrying about optimization of the handlers because I figured it wasn't a big deal to use a bit more resources there in most cases, but I guess when you're printing a ton of these messages, it's going to show up :)

PR #393, opened 2024-08-05
Handle invalid UTF-8 in source code

Porges:
At the moment if invalid UTF-8 is passed to the source code, then the `Report` will panic when rendered. This PR changes it to properly decode UTF-8 and handle the invalid data in the usual way (i.e. they will be replaced with �).

This can require an extra allocation when rendering `Report`, but only for invalid UTF-8, and only for the "context" area, which is usually small (several lines).

This is simply a one-line change, but there are tests to double-check the behaviour of offsets into invalid source. I added some tests for labels into "good" Unicode source and also into "bad" Unicode source. Both seem to behave as expected.

### DIFF
--- a/tests/graphical.rs
+++ b/tests/graphical.rs
@@ -2349,3 +2349,163 @@ Error: oops::my::inner
     assert_eq!(expected, &out);
     Ok(())
 }
+
+#[test]
+fn after_unicode_width() -> Result<(), MietteError> {
+    #[derive(Debug, Diagnostic, Error)]
+    #[error("pointing")]
+    #[diagnostic(code(pointing_at))]
+    struct E {
+        #[label("something of interest")]
+        src: SourceSpan,
+    }
+
+    let unicode_source = "höööt!";
+
+    // make err pointing at the t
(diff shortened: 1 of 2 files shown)


tests/graphical.rs
```diff
@@ -2349,3 +2349,163 @@ Error: oops::my::inner
...
+    };
+
+    let result = fmt_report(Report::new(err).with_source_code(invalid_source));
+
+    let expected = "decode_err
+
+  × decoding error
+   ╭────
+ 1 │ malformed h�XYZ
+   ·             ┬
```
Porges, line 2503: This is correctly pointing at the `X`, although it is hard to tell from the Github preview.

zkat:
what's the deal with clippy. sigh.

waywardmonkeys:
@zkat I've got clippy fixed in PR #395.

zkat:
awesome thanks. You can rebase now.

Porges:
> awesome thanks. You can rebase now.

Fixed the 1.70 build as well. Should be good to go.

zkat (approved):

PR #395, opened 2024-08-06
Fix `clippy::doc_lazy_continuation` lints

waywardmonkeys:


waywardmonkeys:
Hmm, there's other errors beyond just clippy. Let's fix them too.

PR #401, opened 2024-08-16
Inherit source code to causes as well as related errors

Porges:
At the moment outer source code is supplied to inner related errors but not to causes. This PR changes that.

This allows late-supplied source code to be supplied using concrete types. Comparisons below…

## Concrete types

Firstly, if we wrap our inner error with some well-typed outer wrapper:

```rust
#[derive(thiserror::Error, Debug, miette::Diagnostic)]
#[error("inner-error")]
struct InnerError {
    #[label("inner-label")]
    label: (usize, usize),
}

#[derive(thiserror::Error, Debug, miette::Diagnostic)]
#[error("outer-error")]
struct OuterError {
    #[source_code]
    source_code: String,

    #[diagnostic_source]
    inner: InnerError,
}

fn main() -> miette::Result<(), miette::Report> {
    Err(OuterError {
        source_code: "source code".to_string(),
        inner: InnerError { label: (7, 4) },
    })?
}
```

At the moment we don’t get any source information rendered.

Output:

```
Error: 
  × outer-error
  ╰─▶   × inner-error
```

## Wrapped in Report

On the other hand, if we wrap using `miette::Report`, which loses the underlying type information, we get the source code:

```rs
#[derive(thiserror::Error, Debug, miette::Diagnostic)]
#[error("inner-error")]
struct InnerError {
    #[label("inner-label")]
    label: (usize, usize),
}

fn main() -> miette::Result<(), miette::Report> {
    Err(miette::Report::new(InnerError { label: (7, 4) }).with_source_code("source code"))?
}
```

Output:
```
Error: 
  × inner-error
   ╭────
 1 │ source code
   ·        ──┬─
   ·          ╰── inner-label
   ╰────
```

## Using related

Currently we can achieve something using `#[related]` but that is semantically-inappropriate for some use cases:

```rs
#[derive(thiserror::Error, Debug, miette::Diagnostic)]
#[error("inner-error")]
struct InnerError {
    #[label("inner-label")]
    label: (usize, usize),
}

#[derive(thiserror::Error, Debug, miette::Diagnostic)]
#[error("outer-error")]
struct OuterError {
    #[source_code]
    source_code: String,

    #[related]
    inner: Vec<InnerError>,
}

fn main() -> miette::Result<(), miette::Report> {
    Err(OuterError {
        source_code: "source code".to_string(),
        inner: vec![InnerError { label: (7, 4) }],
    })?
}
```

The rendering is also different since they're rendered as adjacent errors rather than causes:

```
Error: 
  × outer-error

Error:
  × inner-error
   ╭────
 1 │ source code
   ·        ──┬─
   ·          ╰── inner-label
   ╰────
```

## With proposed change

After this PR, the output for the first example (exact same code) should be similar to the last example, except exhibiting cause-nesting:

```
Error: 
  × outer-error
  ╰─▶   × inner-error
         ╭────
       1 │ source code
         ·        ──┬─
         ·          ╰── inner-label
         ╰────
```

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -217,9 +217,18 @@ impl GraphicalReportHandler {
         f: &mut impl fmt::Write,
         diagnostic: &(dyn Diagnostic),
     ) -> fmt::Result {
+        self.render_report_inner(f, diagnostic, diagnostic.source_code())
+    }
+
+    fn render_report_inner(
+        &self,
+        f: &mut impl fmt::Write,
+        diagnostic: &(dyn Diagnostic),
+        parent_src: Option<&dyn SourceCode>,
+    ) -> fmt::Result {
+        let src = diagnostic.source_code().or(parent_src);
         self.render_header(f, diagnostic)?;
-        self.render_causes(f, diagnostic)?;
-        let src = diagnostic.source_code();
+        self.render_causes(f, diagnostic, src)?;
         self.render_snippets(f, diagnostic, src)?;
         self.render_footer(f, diagnostic)?;
         self.render_related(f, diagnostic, src)?;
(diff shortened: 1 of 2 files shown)


Porges:
The other solution I had considered was to pull up labels from causes into the parent and merge all of the labels together. I think this would be nice but maybe enabled with a different attribute?

zkat (approved):

PR #403, opened 2024-09-06
feat: add `from_err()` method to `Report`

johnbchron:
There's no nice way to construct a bare `Report` from a regular `Error` type. You can construct `Result<_, Report>` from `Result::<_, E: ...>`, but not from an error by itself.

My workaround where `e: impl Error + Send + Sync + 'static` is to do `Err::<(), _>(e).into_diagnostic().unwrap_err()`, which obviously is ugly.

I added a `Report::from_err()` method that solves this problem in the same way as the `IntoDiagnostic` trait, by using its `DiagnosticError` type.

### DIFF
--- a/src/eyreish/error.rs
+++ b/src/eyreish/error.rs
@@ -425,6 +425,14 @@ impl Report {
         }
         .into()
     }
+
+    /// Construct a [`Report`] directly from an error-like type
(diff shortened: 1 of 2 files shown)


zkat (approved):

zkat:
hey this is nice. Thanks!

PR #407, opened 2024-11-13
examples: add serde_json integration example

Schniz:
I hope that this implementation makes sense. This is what I just added in [fnm](https://github.com/Schniz/fnm)


zkat:
this is great! thank you so much :)

I think this is a super common thing to do, tbh.

PR #409, opened 2024-11-16
feat: Implement `WrapError` for `Option`

### LINKED ISSUE #408: Implement `WrapError` for `Option`
`anyhow` [implements it's `Context` trait on options](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#impl-Context%3CT,+Infallible%3E-for-Option%3CT%3E). This helps avoid some boilerplate:

```rust
use anyhow::Context;

let map = HashMap::from([("a", 1), ("b", 2), ("c", 3)]);
let _ = map.get("d").with_context(|| format!("Key 'd' was missing from map: {map:?}"))?;
```

Currently the best way to do this with `miette` is by using a `let else` then `bail!`:

```rust
use miette::bail;

let map = HashMap::from([("a", 1), ("b", 2), ("c", 3)]);
let Some(_) = map.get("d") else {
    bail!("Key 'd' was missing from map: {map:?}"));
};
```

I'm willing to implement this if desired

jalil-salame:
Implement `WrapError` for `Option<T>`. This is inline with `anyhow` that also implements `Context` for `Option<T>`.

The implementation requires us to introduce a `DisplayError` internal only type, that creates an error from a type that only implements `Display` (`Report::from_adhoc` requires the type to also implement `Debug`, but `WrapError` only requires it to implement `Display`).

For this I copied `MessageError` and adapted it to implement `Debug` using the underlying type's `Display` impl. This is a bit of a hack, but anyhow does [something similar][1].

Closes #408

[1]: https://docs.rs/anyhow/latest/src/anyhow/wrapper.rs.html#34

### DIFF
--- a/src/eyreish/context.rs
+++ b/src/eyreish/context.rs
@@ -38,6 +38,44 @@ mod ext {
     }
 }
 
+impl<T> WrapErr<T, std::convert::Infallible> for Option<T> {
+    fn wrap_err<D>(self, msg: D) -> Result<T, Report>
+    where
+        D: Display + Send + Sync + 'static,
+    {
+        match self {
+            Some(t) => Ok(t),
+            None => Err(Report::from(crate::eyreish::wrapper::DisplayError(msg))),
+        }
+    }
+
+    fn wrap_err_with<D, F>(self, msg: F) -> Result<T, Report>
+    where
+        D: Display + Send + Sync + 'static,
+        F: FnOnce() -> D,
+    {
+        match self {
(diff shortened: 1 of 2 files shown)


jalil-salame:
If you want I can also create a `pub(crate) fn from_display` in `Report` that does the same as `from_adhoc` but without the `Debug` bound.

zkat (approved):

zkat:
thanks for this!

PR #413, opened 2024-11-29
docs: introduce primary param in label

bvanjoi:


JustusFluegel:
Hey quick note: This broke  ci, since this wasn't added in `src/lib.rs` and then a readme was generated using `cargo make readme` but instead directly added to `Readme.md`.

@zkat can we revert this until that is fixed so that main works again?
@bvanjoi I would suggest creating a new pr with the changes in the proper places, maybe? If you don't want to bother with this anymore feel free to tell me then I'll quickly do that for you, but I don't want to take any credit I don't deserve here :)

Thanks :)

PR #415, opened 2024-12-08
docs: fix broken link in README.md

akiomik:


zkat:
This needs to be fixed in src/lib.rs, the README.tpl needs to have the link added, and then you need to run `cargo make readme` to update the readme itself.

akiomik:
Thanks for your review. I've fixed the `README.md` with `cargo make readme`. [The link was  already in the `README.tpl`](https://github.com/zkat/miette/blob/b2011f693b1e912d28e84591b031e5af7af4904b/README.tpl#L6) and did not need to be updated.

zkat (approved):

Issue #416, opened 2024-12-22
Rendering: Provide some way to show multiple nested errors (or display `related` as nested)


elkowar:
Currently, the `related` errors of a given report will show just one after another, without any clear indication of them belonging tightly to the outer error.

I.e., in my project, right now I have errors like these:
![image](https://github.com/user-attachments/assets/2da3c5a1-5ed8-44bd-b9fd-ed0aa108c672)

Which are supposed to express that the two related errors are actually part of the outer error  -- clearly not quite the intended use for `related`.

I can get closer to a more intuitive visualization by storing the related errors in a simple `struct ErrList(#[related] Vec<Report>)`:

![image](https://github.com/user-attachments/assets/45ad8e48-cdcd-4dd6-bf12-a9b9e2447200)

However, this now requires me to have another error _message_ be displayed, which often doesn't add any semantic value, and takes up a lot of space.

For cases like these, where I really want to show several siblings as _nested_ within one outer error, I'd optimally want to have them be displayed roughly like this:

![image](https://github.com/user-attachments/assets/77ebbd88-2bb4-4a26-a091-7553fcbcec9a)

My proposal here would be to add a variant of "related" errors that explicitly declare them as nested, i.e.:

```rs
#[derive(Diagnostic, Error, Debug)]
#[error("{message}")]
#[diagnostic]
struct ManyErrors {
  message: String
  #[related(nested)]
  errors: Vec<Report>
}
```

zkat:
This is a reasonable proposal! I’d be happy to take this change :)

elkowar:
Nice! I'd love to take a stab at this, but I'm a bit unsure on how to approach an implementation here:

If the idea is to actually have this as a property of the related attribute, then I don't know how to express that within the current structure of the Diagnostic trait.
Alternatively, I could just add a separate `nested` field to the Diagnostic trait, which would potentially feel cleaner, but is also a much larger change to the API.
A third, minimally invasive, approach would be to just make the way related errors are shown a global setting in the `GraphicalReportHandler`.

What do you think, @zkat? Or is there something else I'm missing?

zkat:
Probably a configuration in the GraphicalReportHandler is probably best, esp since this is all about controlling how things look, rather than some fundamental property about the relationship between diagnostics

PR #417, opened 2024-12-22
Add show_related_errors_as_nested-option

elkowar:
This PR adds a new option to the `MietteHandlerOpts` that allows rendering related errors as nested, like this:

image

implements #416


zkat (approved):
this is so fancy 🤩

PR #418, opened 2025-01-08
fix(graphical): prevent leading newline when no link/code

unbyte:
#358 introduced a new behavior: if a diagnostic has no `code`, the rendered report will contain a leading newline.

I'm not sure if this was an intentional change or an unintended side effect, but as a long-time miette user, I wouldn't expect my rendered report to include an extra leading newline - whether I'm outputting to the terminal, displaying in web page, or running snapshot tests. So I opened this PR.

Please let me know if I've misunderstood anything.

### DIFF
--- a/src/handlers/graphical.rs
+++ b/src/handlers/graphical.rs
@@ -237,7 +237,7 @@ impl GraphicalReportHandler {
         parent_src: Option<&dyn SourceCode>,
     ) -> fmt::Result {
         let src = diagnostic.source_code().or(parent_src);
-        self.render_header(f, diagnostic)?;
+        self.render_header(f, diagnostic, false)?;
         self.render_causes(f, diagnostic, src)?;
         self.render_snippets(f, diagnostic, src)?;
         self.render_footer(f, diagnostic)?;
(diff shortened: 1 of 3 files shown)


zanieb:
I believe this is the cause of the regression blocking our upgrade in https://github.com/astral-sh/uv/pull/9568 — many snapshots changed :)

zkat:
@zanieb do you think you could run your test suite against this PR and see if it does actually fix the issue? It does seem to be the same thing though but I want to avoid thrash if possible

zkat (approved):
I wanna make sure this fixes uv's upgrade but everything looks good! Thank you for taking the time to do this. Sorry it took a while to get to this :)

zanieb:
Yep this looks good to me https://github.com/astral-sh/uv/pull/10608

zkat:
@zanieb thanks for taking a look!

zanieb:
This bites us every time we update our dependency tree :) would appreciate a release when you have the chance. Let me know if we can help!

zkat:
Ah yes. I’ll roll one out soon

zkat:
@zanieb This is now out as 7.5.0. Pardon the delay. Life has been a thing recently :) haha

zanieb:
No problem. Thank you so much!

Issue #162, opened 2022-04-24
Using default template type for the #[source] field


Finomnis:
I currently have the `source` field of my error enum be `Box<dyn Error + Send + Sync>`.

In my project, I would like to make this `source` field type configurable with a generic with `Box<dyn Error + Send + Sync>` as the default type, but I can't seem to find the correct combination of restrictions for the generic.

So, basically the TLDR; question is: What should be written in the `<???>` to make this code compile?

```rust
#[derive(Debug, thiserror::Error, miette::Diagnostic)]
pub enum MyError<ErrType: <???> = Box<dyn Error + Send + Sync>> {
    #[error("Error containing another error")]
    AnError(#[source] ErrType),
}
```

I have a more detailed discussion about my attemps and problems I encountered [here on stackoverflow](https://stackoverflow.com/questions/71980268/miette-how-to-specify-errors-that-contain-a-generic-that-defaults-to-boxdyn-e).

Thanks for the great work :) <3

zkat:
Hmmm.

This seems to be a limitation of the `Diagnostic` macro, but I'm not sure how to work around it. To show something that works, for example:

```rust
    #[derive(Debug, Error)]
    enum MyError<ErrType = Box<dyn std::error::Error + Send + Sync>> {
        #[error("Error containing another error")]
        AnError(#[source] ErrType),
    }

    impl<ErrType: 'static + std::error::Error> Diagnostic for MyError<ErrType> {
    }
```

This will compile just fine! And if you feel up to implementing `miette::Diagnostic` by hand, this is more than good enough :)

I'm not actually sure what else can be done here, besides some clever wrapping of types that I haven't quite thought through yet, but in essence, the issue is that the `Diagnostic` proc macro, while it's _aware_ of generics, doesn't magically know that you need to add that extra `'static + std::error::Error` bound to it (or which type to add it to), and if you end up adding the trait bound to the original enum definition, you'll run into a conflict because `Box<dyn Error>` doesn't implement `std::error::Error`, as you pointed out.

So the immediate workaround to unblock you is to stop using the derive macro for now. As to a more robust solution that lets you use the derive macro... I have no idea, but I'm open to help/suggestions from folks who have more brain cells for this than I do haha.

jdonszelmann:
This  could be solved with some way to explicitly give the bounds on T through an attribute, a bit like how with `async_trait` you can add the `?Send` bound:
https://docs.rs/async-trait/latest/async_trait/#non-threadsafe-futures

PR #421, opened 2025-02-02
Add required bounds to derived impl

JustusFluegel:
This is an attempt to make the derive macro for `Diagnostic` work better with generics, allowing to write code like for example

```rust
#[derive(Debug, thiserror::Error, miette::Diagnostic)
enum MyError<T> {
     #[error(transparent)]
     #[diagnostic(transparent)]
     Other(#[from] T).
     #[error("bar")]
     #[diagnostic(help = "baz")]
     Foo
}
```

This should for example make #162 just work, as well as things like

```rust
#[derive(Debug, thiserror::Error, miette::Diagnostic)
struct MyOtherError<T> {
     #[label]
     label: T
}
``` 

and the other supported attributes.

AFAIK the only gotcha with this implementation right now is that `#[related]` only works with concrete collection types and not with generic collections, but the type inside the collection can be generic. (meaning: `Vec<T>` works but not any arbitrary `C` where `C: IntoIter<Item: Diagnostic>`), and I think that is a current limitation of the rust compiler / the current Diagnostic trait design.

Questions: Do we want to gate this behind a feature flag since it pulls in extra-traits from syn and performs a bunch of extra work?

### DIFF
--- a/miette-derive/Cargo.toml
+++ b/miette-derive/Cargo.toml
@@ -13,4 +13,4 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.78"
 quote = "1.0.35"
-syn = "2.0.48"
+syn = { version = "2.0.48", features = ["extra-traits"] }
--- a/miette-derive/src/diagnostic.rs
+++ b/miette-derive/src/diagnostic.rs
@@ -11,6 +11,7 @@ use crate::label::Labels;
 use crate::related::Related;
 use crate::severity::Severity;
 use crate::source_code::SourceCode;
+use crate::trait_bounds::TraitBoundStore;
 use crate::url::Url;
 
 pub enum Diagnostic {
@@ -19,11 +20,13 @@ pub enum Diagnostic {
         ident: syn::Ident,
         fields: syn::Fields,
         args: DiagnosticDefArgs,
+        bound_store: TraitBoundStore,
     },
     Enum {
         ident: syn::Ident,
         generics: syn::Generics,
         variants: Vec<DiagnosticDef>,
+        bound_store: TraitBoundStore,
     },
 }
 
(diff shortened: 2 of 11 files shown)


JustusFluegel:
This should be mostly good, I'll add some tests but I'd appreciate a comment-type review in the meantime if possible :)

JustusFluegel:
tests are now there as well, I am sure there are some things I should change since this is a pretty big pr and I am not that familiar with this project but from my perspective this should be finished apart from that.

zkat:
Putting it behind a feature flag seems wise, yes.

JustusFluegel:
Hi, as I am in the process of moving right now I will probably only get back to this the next weekend or so, but my current plan for this is:

- Add the suggested feature flag ( I would probably name it `perfect_derive` which [aligns](https://smallcultfollowing.com/babysteps/blog/2022/04/12/implied-bounds-and-perfect-derive/) with other work on the same idea across the ecosystem )
    - Since we don't add bounds to the impl at all right now this should be a purely additive feature, which is what all feature flags should be
- Write a higher level summary on the changes of this pr
    - explicitly list all the codegen changes inside the derive macro ( to make it easier to spot accidental breaking changes in review, even though I didn't mean to introd

JustusFluegel:
( accidentally pressed the close button, sorry for the noise )

JustusFluegel:
Hey just a quick update that I am still working on this, just life kinda got in the way and I didn't have as much time as I wished to work on this. Will try to get this to a review-able state soon tho :)

JustusFluegel:
So a functional summary of the changes introduced in this pr:

The general idea of perfect-derive is to keep track of the types of fields of structs and enums and look at the context what they are used for inside the derive macro and then add trait bounds (additional where clause items) to the types automatically.

To facilitate this, this pr introduces a `TypeParamBoundStore` which is basically a map from types (of fields) to trait bounds which will then be added to the where clause at the end when generating the trait impl. 
Additionally, the `TypeParamBoundStore` implements some simple heuristics to determine if trait bounds added to it via it's methods are actually generic, e.g. have the generics used inside of them.
This prevents adding trivial type bounds like `String: Display` to the where clause, which while not breaking (the compiler accepts such trivial bounds) are unnecessary.  It also ensures no duplicate bounds are added to each type.

At the point of implementation, where the struct fields are then used, we always add the where bounds to the store as if the type is generic, meaning for example `miette::Diagnostic` for `#[diagnostic(transparent)]`. 

When the `perfect_derive` feature is disabled, the `TypeParamBoundStore` is then just replaced with a empty mock implementation where it is a ZST, which should be completely optimized away by the compiler.

Using the `TypeParamBoundStore` at the place of use required passing a `&mut` reference to it around, which is why there are quite a few changes related to that in this pr.

JustusFluegel:
I hope the last 2 comments help a bit with reviewing this pr, I know it is not super small (~+600 lines overall) but I appreciate if you could take the time :)

Thank you in advance
Justus Flügel

JustusFluegel:
One Idea I had to make this strictly non-breaking is to continue to emit the old declarations in the one place where it could be seen as breaking if the feature is not enabled

JustusFluegel:
Also: May I kindly request a ci re-run? :)

JustusFluegel:
Quick ping @zkat just if you have seen this? No worries if you currently don't have the time/will/... to look at this, I'd appreciate a quick comment that you aware of this so that I know that it isn't any issue of visibility tho  :)

JustusFluegel:
Rebased on top of `main` :)

PR #425, opened 2025-02-19
feat(highlighting): implement syntax highlighting for unnamed source codes

glevco:
Currently, both `name` and `language` are part of the `SpanContents` abstraction, which comes from the `NamedSource` struct. It would make more semantic sense if those were part of the `SourceCode` abstraction instead. This would allow us to use syntax highlighting in unnamed source codes, which is not currently possible. This PR proposes this.

### Changes

- Rename `NamedSource` to `MietteSourceCode`, adding `language` to it.
- Move `name()` and `language()` from `SpanContents` trait to `SourceCode` trait.
- Move `name` and `language` from `MietteSpanContents` struct to `MietteSourceCode` struct.

This is a breaking change because it refactors public APIs.

#### Before

```rust
let src = NamedSource::new("bad_file", "fn hello() {}").with_language("Rust");
let report = miette!("oops!").with_source_code(src);
```

#### After

```rust
let src = MietteSourceCode::new("fn hello() {}").with_name("bad_file").with_language("Rust");
let report = miette!("oops!").with_source_code(src);

// unnamed with language, not possible before!
let src = MietteSourceCode::new("fn hello() {}").with_language("Rust");
let report = miette!("oops!").with_source_code(src);
```


zkat (commented):
I generally like this, but I'm gonna have to think more about moving those specific things around. I think there might be other things to be done about the struct to-be-formerly-known-as-`NamedSource` but I'll try and remember what else I thought might be nice.

I also don't know when I'll want to have a breaking release of miette. I try to keep them very very sparse, so if you can think of a way to do this _without_ the breaking change, even if we eventually do the "nice" version, then that'll get unnamed source code highlighting working sooner for the community.

glevco:
@zkat sure, I can do that. I'll simply create the `MietteSourceCode` like in this PR, but as a new struct instead of replacing the `NamedSource` with it. Then you can deprecate `NamedSource` and eventually remove it.

src/protocol.rs
```diff
@@ -546,12 +511,6 @@ impl<'a> SpanContents<'a> for MietteSpanContents<'a> {
     fn line_count(&self) -> usize {
         self.line_count
     }
```
cgettys-microsoft, line 513: I'm not sure this is optimal to move off of SpanContents for reasons I noted in #450.

Specifically, today, the interface could handle fun scenarios like markdown like this:
```markdown
# An introduction to Rust
Here's a Rust program!
```rust
fn hello_world()
{
   println!("hello world")
}
```
```
Language being a property of SpanContents, IMO, thus makes sense. you can have multiple languages in one "SourceCode"
If you don't like the example of markdown, consider Razor/Blazor stuff, HTML, JavaScript, and CSS frequently living together, and so on and so on. Yes, the new interface could do this too, but it'd require representing those parts of the file as separate SourceCode.

I do see that there could be a more convenient interface; but I'm not (personally, for whatever little my opinion might be worth) sure this new interface is better.

PR #426, opened 2025-02-27
update `thiserror` from 1.0.56 to 2.0.11

nogiro:
Hello,

I'm using `watchexec`, which depends on `miette` and `thiserror 2.x`. `miette` currently depends on `thiserror 1.x`, which results in both versions being included in the dependency tree.

I ran `cargo test`, but I'm not sure if there's anything else I should check. Please let me know if there's anything else I need to do.

Thanks!


nogiro:
I'm fixing for the CI `Minimal versions check`. Please wait...

nogiro:
I fixed it.

PR #436, opened 2025-04-26
Remove `thiserror` from dependencies

### LINKED ISSUE #435: Remove `thiserror` dependency?
This issue is somewhat related to #433 and #434 — `thiserror` is a non-optional dependency that's dragging in the `syn` crate. It should be possible to manually implement `Error` for the couple of structs currently using it outside of testing code.

Obviously it's useful to keep as a `dev-dependency`, so I'll try to create a PR that just makes the minimal changes needed to get it out of the main dependency list!

TheLostLambda:
Closes #435 and should just be a patch-level SemVer change!

Currently blocking the removal of `syn` from `kdl-rs`s dependency tree, but with this PR merged, that should become rather straightforward!

I also found and fixed a bug leading to compilation errors in `ffe374c`, some clippy lints, and reused some code in `53c8330`.

I also added tests wherever I was replacing `Error` or `Diagnostic` derives with something hand-written, just to be sure I wasn't changing any behavior!

### DIFF
--- a/src/eyreish/error.rs
+++ b/src/eyreish/error.rs
@@ -280,7 +280,7 @@ impl Report {
     /// The root cause is the last error in the iterator produced by
     /// [`chain()`](Report::chain).
     pub fn root_cause(&self) -> &(dyn StdError + 'static) {
-        self.chain().last().unwrap()
+        self.chain().next_back().unwrap()
     }
 
     /// Returns true if `E` is the type held by this error object.
--- a/src/eyreish/into_diagnostic.rs
+++ b/src/eyreish/into_diagnostic.rs
@@ -1,12 +1,20 @@
-use thiserror::Error;
+use std::{error::Error, fmt::Display};
 
 use crate::{Diagnostic, Report};
 
 /// Convenience [`Diagnostic`] that can be used as an "anonymous" wrapper for
 /// Errors. This is intended to be paired with [`IntoDiagnostic`].
-#[derive(Debug, Error)]
-#[error(transparent)]
+#[derive(Debug)]
 pub(crate) struct DiagnosticError(pub(crate) Box<dyn std::error::Error + Send + Sync + 'static>);
+
+impl Display for DiagnosticError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let msg = &self.0;
+        write!(f, "{msg}")
+    }
+}
+impl Error for DiagnosticError {}
+
 impl Diagnostic for DiagnosticError {}
 
 /**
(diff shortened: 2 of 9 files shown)


TheLostLambda:
Not sure what the doc consistency check is about, @zkat ! It seems to generate this diff, deleting a section of the README if I run the command in the makefile, but I don't know if that's correct...

image

src/eyreish/into_diagnostic.rs
```diff
@@ -1,12 +1,20 @@
...
+#[derive(Debug)]
 pub(crate) struct DiagnosticError(pub(crate) Box<dyn std::error::Error + Send + Sync + 'static>);
+
+impl Display for DiagnosticError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let msg = &self.0;
+        write!(f, "{msg}")
+    }
+}
+impl Error for DiagnosticError {}
```
zkat, line 16: `error(transparent)` forwards the `source` method

src/eyreish/error.rs
```diff
@@ -280,7 +280,7 @@ impl Report {
     /// The root cause is the last error in the iterator produced by
     /// [`chain()`](Report::chain).
     pub fn root_cause(&self) -> &(dyn StdError + 'static) {
-        self.chain().last().unwrap()
+        self.chain().next_back().unwrap()
```
zkat, line 283: For why?

zkat (changes requested):
So exciting! Thanks so much for taking the time!

src/eyreish/into_diagnostic.rs
TheLostLambda: Ach! Good catch! I think my tests must have missed that by not using an error with a source... I'll fix that up now!

TheLostLambda: Should be fixed now!

TheLostLambda:
I suppose changing MSRV is probably not a SemVer patch change... I'll roll back my fancy-pants `io::Error::other(...)` from 1.74 :sweat_smile: ...

TheLostLambda:
Trying to fix that README thing now too, since the CI bullied me an I'm looking for revenge (green checkmark...)

TheLostLambda:
CI successfully slain :crossed_swords: :relieved:

Should be ready for review again!

src/eyreish/error.rs
zkat: Oh! Huh! Good call

zkat (approved):

zkat:
Thanks a bunch for this! I'll roll a new release later today. Feel free to ping me if I forget :)

TheLostLambda:
Sounds like a great plan! That'll be great for unblocking the `kdl-rs` changes :relaxed:

PR #450, opened 2025-09-25
Add MietteSpanContents::with_name, improve docs, fix tests being compiled in downstream libraries, fix NamedSource's Debug impl

### LINKED ISSUE #449: `Debug` implementation for `NamedSource` missing call to `.finish()`
In the `Debug` implementation for `NamedSource`

```rust
impl<S: SourceCode> std::fmt::Debug for NamedSource<S> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("NamedSource")
            .field("name", &self.name)
            .field("source", &"<redacted>")
            .field("language", &self.language);
        Ok(())
    }
}
```
https://github.com/zkat/miette/blob/main/src/named_source.rs#L13-L21

the [`.finish()`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html#method.finish) method is not called, which causes it to not close the struct, resulting in confussing output such as `MyError { source: NamedSource { name: "foo.txt", source: "<redacted>", language: None, other_fields_in_my_error: usize }`

cgettys-microsoft:
* Add a 2-line utility method to MietteSpanContents - MietteSpanContents::with_name. You could easily do the same thing anyway, but it'd be pretty ugly today - make it easy.
* Fix some tests under src/ not being gated under ```#[cfg(test)]```. Yes, LTO would eliminate them at link time most likely, but still, may improve compile times downstream a bit, and definitely better cleanliness wise unless I'm missing something.  I did add modules to contain the tests, but I can undo that part if you rather - the ```#[cfg(all(test, feature = ..)]``` bit is enough.
* Improve discoverability of how SourceCode actually communicates the file name and language. This is well documented in the lib.rs docs on the index, but not super discoverable if you miss it there. I came across the comment on NamedSource that says SourceCode provides the name, and then looked at SourceCode looking for a trait method or something, and found nothing. So document in both those places. To be clear, I think the design is good - as it allows a SourceCode implementation to provide different names or languages for different spans, which can be useful in scenarios like a .md file containing a code block in another language - it just wasn't clear to me at first.
* Also close #449 - call ```finish```, replace the ```Ok(())``` with returning the result of ```finish```

Happy to split out this PR into 4 PRs if you prefer. Thanks for the awesome library!

### DIFF
--- a/miette-derive/src/forward.rs
+++ b/miette-derive/src/forward.rs
@@ -150,7 +150,7 @@ impl Forward {
                 Self::#variant { #field_name, .. } => #field_name.#method_call,
             },
             Forward::Unnamed(index) => {
-                let underscores: Vec<_> = core::iter::repeat(quote! { _, }).take(*index).collect();
+                let underscores: Vec<_> = std::iter::repeat_n(quote! { _, }, *index).collect();
                 let unnamed = format_ident!("unnamed");
                 quote! {
                     Self::#variant ( #(#underscores)* #unnamed, .. ) => #unnamed.#method_call,
--- a/src/highlighters/syntect.rs
+++ b/src/highlighters/syntect.rs
@@ -125,7 +125,7 @@ impl HighlighterState for SyntectHighlighterState<'_> {
                 line,
                 &self.highlighter,
             )
-            .map(|(style, str)| (convert_style(style, use_bg_color).style(str)))
+            .map(|(style, str)| convert_style(style, use_bg_color).style(str))
             .collect()
         } else {
             vec![Style::default().style(line)]


zkat (approved):

cgettys-microsoft:
I can't reproduce any of these failures either on main or on my branch on 1.88. The enum does look unused though.

I'll see if I can get rustup to install 1.70 and see if I can work out what's going on

cgettys-microsoft:
Ah wait, it's 1.90, not 1.70. Got it. Looks to be pre-existing, but let me see if I can lend a hand anyway :)

cgettys-microsoft:
This is getting strange.... fixing the "unnecessary" parantheses results in:
`
warning: error finalizing incremental compilation session directory `\\?\REPO_LOCATION_REMOVED\miette\target\debug\incremental\miette-0e64h1vlgbg44\s-hbgqkx5vnp-1q0dav7-working`: Access is denied. (os error 5)
`

cgettys-microsoft:
I fixed a few of the errors in PR #451

The one that happens if I remove the "unnecessary" parentheses remains baffling to me. I'm assuming it's a proc macro issue even though it's not giving a nice error message, and that the proc macro is not parsing expressions it needs to parse unless there are parens.
I just... can't quite work out where in the proc macro the issue is if so. I can of course just suppress the warning I guess... but that's unsatisfying.
Edit: looks like Windows Defender was just being uncooperative with certain Rustc versions... fun

cgettys-microsoft:
This is now stacked on top of #451 and should pass CI, finally (pass on my fork: https://github.com/cgettys-microsoft/miette/actions/runs/18043812032). But given that #451 bumps MSRV and the like, that should probably be reviewed prior to merging this, and then this rebased? Up to you :)

PR #451, opened 2025-09-26
fix(ci, clippy): raise MSRV to 1.82, fix rust 1.90 warnings

cgettys-microsoft:
Addresses a few new warnings in 1.90.

Update MSRV to 1.82; even adding --locked, I just couldn't get CI to pass with the declared versions and 1.70. If someone else can, obviously that'd be a better solution.


cgettys-microsoft:
Still working on the MSRV related errors.

cgettys-microsoft:
I tried adding --locked, but Rust 1.70.0 still wouldn't compile:
https://github.com/cgettys-microsoft/miette/actions/runs/18043280615/job/51347678859

1.76 wasn't enough either, had to go up to 1.82 (which to be fair is still over a year old) to get this to pass in CI.

Issue #443, opened 2025-05-17
no_std support


lucascool12:
Hi cool library. I'm wondering if there is any interest in making the core of this crate `no_std`?
This would make it usable on places where `std` is not available or undesired.
There seem to be some forks who have done this one way or the other: <https://github.com/0xMiden/miette/tree/no-std> which is forked from <https://github.com/bitwalker/miette/tree/no-std>.

It would be quite nice if this work could somehow be integrated into the original library!

Some notes: doing this will bump the msrv to 1.81.0 since the `Error` trait is only available on a stable toolchain since then: [source](https://doc.rust-lang.org/core/error/trait.Error.html).

bitwalker:
Hi, I'm the one who made that fork. At the time, I think there was some movement towards implementing no-std support, but it hadn't been done yet, and I needed it quick-fast, so I forked and made the changes necessary.

With the release of thiserror 2.0, and the stabilization of the `error_in_core` feature, I think it would be quite straightforward to port the changes from 0xMiden/miette into mainline and do a release with a higher MSRV. The actual changes themselves are mostly about using liballoc instead of libstd for heap-allocated types, and gating anything that actually requires libstd behind a `std` feature, rather than anything particularly invasive.

I'd certainly prefer to upstream my work rather than maintain a fork, but implementing no-std support does place extra maintenance burden on the maintainers, so it is really up to them on whether they are open to that.

cgettys-microsoft:
#451 (not released yet) raised MSRV to 1.82, so the 1.81.0 MSRV bit isn't a problem anymore, FWIW. 
I'm not one of the maintainers, so I can't speak to openness to no-std support - that's a question better asked of @zkat I think.

PR #455, opened 2025-10-21
feat: no-std support

huitseeker:
Adds support for no-std.

This is a slight modernization of the miette fork by @bitwalker mentioned in #443, found at https://github.com/bitwalker/miette/compare/main...no-std

All the credit for the contribution goes to @bitwalker, I'm just doing janitorial work around the edges.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,21 +32,18 @@ syntect = { version = "5.1.0", optional = true }
 
 [dev-dependencies]
 semver = "1.0.21"
-
-[build-dependencies]
-rustc_version = "0.2"
-
-# Eyre devdeps
 futures = { version = "0.3", default-features = false }
 indenter = "0.3.3"
 rustversion = "1.0"
 trybuild = { version = "1.0.89", features = ["diff"] }
 syn = { version = "2.0.87", features = ["full"] }
 regex = "1.10"
-
 serde = { version = "1.0.196", features = ["derive"] }
 serde_json = "1.0.113"
 strip-ansi-escapes = "0.2.0"
+
+[build-dependencies]
+rustc_version = "0.2"
 
 [features]
 default = ["derive", "std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["images/", "tests/", "miette-derive/"]
 miette-derive = { path = "miette-derive", version = "=7.6.0", optional = true }
 unicode-width = "0.2.0"
 cfg-if = "1.0.0"
-spin = { version = "0.9", default-features = false, features = ["once"], optional = true }
+spin = { version = "0.9", default-features = false, features = ["once"] }
 
 owo-colors = { version = "4.0.0", optional = true }
 textwrap = { version = "0.16.0", default-features = false, features = ["unicode-linebreak", "unicode-width"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ fancy-no-backtrace = [
     "dep:supports-unicode",
 ]
 fancy = ["fancy-no-backtrace", "dep:backtrace", "dep:backtrace-ext"]
-syntect-highlighter = ["fancy-no-backtrace", "dep:syntect"]
+syntect-highlighter = ["fancy-no-backtrace", "dep:syntect", "std"]
 
 [workspace]
 members = ["miette-derive"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,9 @@ exclude = ["images/", "tests/", "miette-derive/"]
 
 [dependencies]
 miette-derive = { path = "miette-derive", version = "=7.6.0", optional = true }
-unicode-width = { version = "0.2.0", default-features = false }
+unicode-width = "0.2.0"
 cfg-if = "1.0.0"
-# spin is needed for no_std environments (provides Once without std)
-spin = { version = "0.9", default-features = false, features = ["once"] }
+spin = { version = "0.9", default-features = false, features = ["once"], optional = true }
 
 owo-colors = { version = "4.0.0", optional = true }
 textwrap = { version = "0.16.0", default-features = false, features = ["unicode-linebreak", "unicode-width"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,8 +42,6 @@ serde = { version = "1.0.196", features = ["derive"] }
 serde_json = "1.0.113"
 strip-ansi-escapes = "0.2.0"
 
-[build-dependencies]
-rustc_version = "0.2"
 
 [features]
 default = ["derive", "std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,11 @@ rust-version = "1.82.0"
 exclude = ["images/", "tests/", "miette-derive/"]
 
 [dependencies]
+thiserror = "2.0.11"
 miette-derive = { path = "miette-derive", version = "=7.6.0", optional = true }
-unicode-width = "0.2.0"
+unicode-width = { version = "0.2.0", default-features = false }
 cfg-if = "1.0.0"
+spin = { version = "0.9", default-features = false, features = ["mutex", "spin_mutex", "lazy"] }
 
 owo-colors = { version = "4.0.0", optional = true }
 textwrap = { version = "0.16.0", default-features = false, features = ["unicode-linebreak", "unicode-width"], optional = true }
@@ -29,8 +31,10 @@ serde = { version = "1.0.196", features = ["derive"], optional = true }
 syntect = { version = "5.1.0", optional = true }
 
 [dev-dependencies]
-thiserror = "2.0.11"
 semver = "1.0.21"
+
+[build-dependencies]
+rustc_version = "0.2"
 
 # Eyre devdeps
 futures = { version = "0.3", default-features = false }
@@ -45,7 +49,8 @@ serde_json = "1.0.113"
 strip-ansi-escapes = "0.2.0"
 
 [features]
-default = ["derive"]
+default = ["derive", "std"]
+std = ["thiserror/std", "fancy-no-syscall"]
 derive = ["dep:miette-derive"]
 no-format-args-capture = []
 fancy-base = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ rust-version = "1.82.0"
 exclude = ["images/", "tests/", "miette-derive/"]
 
 [dependencies]
-thiserror = "2.0.11"
 miette-derive = { path = "miette-derive", version = "=7.6.0", optional = true }
 unicode-width = { version = "0.2.0", default-features = false }
 cfg-if = "1.0.0"
-spin = { version = "0.9", default-features = false, features = ["mutex", "spin_mutex", "lazy"] }
+# spin is needed for no_std environments (provides Once without std)
+spin = { version = "0.9", default-features = false, features = ["once"] }
 
 owo-colors = { version = "4.0.0", optional = true }
 textwrap = { version = "0.16.0", default-features = false, features = ["unicode-linebreak", "unicode-width"], optional = true }
@@ -31,6 +31,7 @@ serde = { version = "1.0.196", features = ["derive"], optional = true }
 syntect = { version = "5.1.0", optional = true }
 
 [dev-dependencies]
+thiserror = "2.0.11"
 semver = "1.0.21"
 futures = { version = "0.3", default-features = false }
 indenter = "0.3.3"
@@ -45,7 +46,7 @@ strip-ansi-escapes = "0.2.0"
 
 [features]
 default = ["derive", "std"]
-std = ["thiserror/std", "fancy-no-syscall"]
+std = []
 derive = ["dep:miette-derive"]
 no-format-args-capture = []
 fancy-base = [
@@ -57,13 +58,14 @@ fancy-no-syscall = [
 ]
 fancy-no-backtrace = [
     "fancy-base",
+    "std",
     "dep:terminal_size",
     "dep:supports-hyperlinks",
     "dep:supports-color",
     "dep:supports-unicode",
 ]
 fancy = ["fancy-no-backtrace", "dep:backtrace", "dep:backtrace-ext"]
-syntect-highlighter = ["fancy-no-backtrace", "dep:syntect", "std"]
+syntect-highlighter = ["fancy-no-backtrace", "dep:syntect"]
 
 [workspace]
 members = ["miette-derive"]
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,0 @@
-fn main() {
-    // track_caller is always available with our MSRV (Rust ≥ 1.82)
-
-    // Add check-cfg for conditional configurations
-    println!("cargo:rustc-check-cfg=cfg(doc_cfg)");
-}
--- a/build.rs
+++ b/build.rs
@@ -1,15 +1,8 @@
-use rustc_version::{version_meta, Channel};
-
 fn main() {
-    if let Channel::Nightly = version_meta().unwrap().channel {
-        println!("cargo:rustc-cfg=nightly")
-    }
-
     // track_caller is stable since Rust 1.46 (2020), so no version check needed
     println!("cargo:rustc-cfg=track_caller");
 
     // Add check-cfg for conditional configurations
     println!("cargo:rustc-check-cfg=cfg(doc_cfg)");
     println!("cargo:rustc-check-cfg=cfg(track_caller)");
-    println!("cargo:rustc-check-cfg=cfg(nightly)");
 }
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,6 @@
 fn main() {
-    // track_caller is stable since Rust 1.46 (2020), so no version check needed
-    println!("cargo:rustc-cfg=track_caller");
+    // track_caller is always available with our MSRV (Rust ≥ 1.82)
 
     // Add check-cfg for conditional configurations
     println!("cargo:rustc-check-cfg=cfg(doc_cfg)");
-    println!("cargo:rustc-check-cfg=cfg(track_caller)");
 }
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,19 @@
+use rustc_version::{version_meta, Channel};
+
+fn main() {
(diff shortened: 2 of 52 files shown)


miette-derive/src/code.rs
```diff
@@ -56,13 +56,13 @@ impl Code {
                 let code = &code.as_ref()?.0;
                 Some(match fields {
                     syn::Fields::Named(_) => {
-                        quote! { Self::#ident { .. } => std::option::Option::Some(std::boxed::Box::new(#code)), }
```
cgettys-microsoft, line 59: why isn't Option qualified after this change? just curious

src/error.rs
```diff
@@ -66,42 +77,58 @@ impl Diagnostic for MietteError {
...
     }
 
+    #[cfg(feature = "std")]
     impl Error for TestError {
         fn source(&self) -> Option<&(dyn Error + 'static)> {
             Some(&self.0)
         }
     }
 
+    #[cfg(not(feature = "std"))]
```
cgettys-microsoft, line 117: Curiosity: why not a different struct, instead of using one in two configs.
If the answer is "won't compile", then this is infectious / not additive, no?

src/eyreish/context.rs
```diff
@@ -29,6 +34,7 @@ mod ext {
     }
 
     impl Diag for Report {
+        #[cfg_attr(track_caller, track_caller)]
```
cgettys-microsoft: is track_caller needed for the no_std implementation, or is that a separate change?

src/eyreish/error.rs
```diff
@@ -816,8 +817,16 @@ impl AsRef<dyn StdError> for Report {
     }
 }
 
+#[cfg(feature = "std")]
 impl std::borrow::Borrow<dyn Diagnostic> for Report {
     fn borrow(&self) -> &(dyn Diagnostic + 'static) {
         self.as_ref()
     }
 }
+
```
cgettys-microsoft: WHy doesn't std::core::Borrow work in std contexts? why two impls, in other words?

src/eyreish/into_diagnostic.rs
```diff
@@ -1,14 +1,17 @@
-use std::{error::Error, fmt::Display};
+extern crate alloc;
+
+use crate::StdError as Error;
```
cgettys-microsoft: why re-name StdError to Error here, but not elsewehre?

src/eyreish/kind.rs
```diff
@@ -45,11 +45,20 @@
 //     let error = $msg;
 //     (&error).miette_kind().new(error)
 
+#[cfg(not(feature = "std"))]
```
cgettys-microsoft: can all these repeated declarations live in a single imported prelude instead, or nah?

src/eyreish/mod.rs
```diff
@@ -81,18 +91,34 @@ impl Diagnostic for InstallError {}
...
+    hook(error)
 }
 
 #[cfg_attr(track_caller, track_caller)]
-#[cfg_attr(not(track_caller), allow(unused_mut))]
-fn capture_handler(error: &(dyn Diagnostic + 'static)) -> Box<dyn ReportHandler> {
-    let hook = HOOK.get_or_init(|| Box::new(get_default_printer)).as_ref();
+pub(crate) fn capture_handler_with_location(
+    error: &(dyn Diagnostic + 'static),
+) -> Box<dyn ReportHandler> {
```
cgettys-microsoft, line 121: Why the Once? can't it just be OnceLock?
I'm having a hard time following what this change is achieving, which means more comments might be needed here. I don't see why two different once's are needed.

src/eyreish/wrapper.rs
```diff
@@ -107,16 +110,6 @@ impl StdError for BoxedError {
     fn source(&self) -> Option<&(dyn StdError + 'static)> {
         self.0.source()
     }
-
```
cgettys-microsoft, line 110: This is a breaking change, no?

src/eyreish/wrapper.rs
```diff
@@ -216,8 +209,16 @@ impl<C> StdError for WithSourceCode<Report, C> {
...
-#[cfg(test)]
+#[cfg(all(test, feature = "std"))]
 mod tests {
+    #[cfg(feature = "fancy")]
+    use std::format;
+    use std::{
+        boxed::Box,
+        string::{String, ToString},
+    };
+    #[cfg(feature = "fancy")]
```
cgettys-microsoft, line 220: shouldn't many of these still be alloc:: qualified types instead, in case someone wants to make fancy no-std too in future?

src/handler.rs
```diff
@@ -491,21 +507,27 @@ mod syscall {
     #[inline]
     pub(super) fn supports_color() -> bool {
         cfg_if! {
-            if #[cfg(feature = "fancy-no-syscall")] {
+            if #[cfg(feature = "fancy-no-backtrace")] {
+                supports_color::on(supports_color::Stream::Stderr).is_some()
+            } else if #[cfg(feature = "fancy-no-syscall")] {
+                // In no-std environment without color support, default to no color support
```
cgettys-microsoft: comment not quite true. fancy-no-syscall can be enabled w/o no-std, right?

src/handler.rs
```diff
@@ -491,21 +507,27 @@ mod syscall {
     #[inline]
     pub(super) fn supports_color() -> bool {
         cfg_if! {
-            if #[cfg(feature = "fancy-no-syscall")] {
+            if #[cfg(feature = "fancy-no-backtrace")] {
```
cgettys-microsoft: this reordering seems like it might be a behavior change.
Shouldn't it be something like:
if fancy-no-syscall, false
maybe else if no-std, false
else if fancy-no-backtrace, check
maybe else, false?
These negated features are so hard to think through. Feels like some of them should be flipped next time there's a major version?

src/handler.rs
```diff
@@ -491,21 +507,27 @@ mod syscall {
...
         }
     }
 
     #[inline]
     pub(super) fn supports_color_has_16m() -> Option<bool> {
         cfg_if! {
-            if #[cfg(feature = "fancy-no-syscall")] {
-                None
-            } else {
+            if #[cfg(feature = "fancy-no-backtrace")] {
```
cgettys-microsoft: again, seems like this may change behavior in with std scenarios?

src/handlers/theme.rs
```diff
@@ -69,12 +71,20 @@ impl GraphicalTheme {
 
 impl Default for GraphicalTheme {
     fn default() -> Self {
-        match std::env::var("NO_COLOR") {
-            _ if !std::io::stdout().is_terminal() || !std::io::stderr().is_terminal() => {
-                Self::none()
+        #[cfg(feature = "fancy-no-syscall")]
```
cgettys-microsoft: this seems ok, though again, the negated features are really tricky...

src/highlighters/mod.rs
```diff
@@ -78,21 +83,28 @@ impl MietteHighlighter {
...
-    #[cfg(feature = "syntect-highlighter")]
     fn default() -> Self {
-        use std::io::IsTerminal;
-        match std::env::var("NO_COLOR") {
-            _ if !std::io::stdout().is_terminal() || !std::io::stderr().is_terminal() => {
-                //TODO: should use ANSI styling instead of 24-bit truecolor here
-                Self(Arc::new(SyntectHighlighter::default()))
+        #[cfg(all(feature = "syntect-highlighter", not(feature = "fancy-no-syscall")))]
+        {
+            use std::io::IsTerminal;
```
cgettys-microsoft, line 89: It looks like there's some deduplication that should be done vs src/handlers/theme.rs
It's already pretty complicated and presumably the two should be kept in sync.
Maybe a helper function?  Ideally that memo-izes it to avoid many redundant syscalls if necessary

Note: the double or triple negation in the old version was making my head hurt, so I made it exhaustive instead
```rust
enum StyleDetection
{
  NoColor,
  Color
}
fn detect_color_support() ->
{
    #[cfg(any(feature = "fancy-no-syscall", not(feature = "std")))]
    { 
      StyleDetection::NoColor
    }
    
    #[cfg(all(not(feature = "fancy-no-syscall"), feature = "std"))]
    { 
        match std::env::var("NO_COLOR")  {
          if !std::io::stdout().is_terminal() || !std::io::stderr().is_terminal() => {
            // TODO: address todo by making this StyleDetection::NoColor?
            //TODO: should use ANSI styling instead of 24-bit truecolor here
            StyleDetection::Color
          }
          Ok(string) => 
          {
            if string == "0" 
            {
              // NO_COLOR = "0" means enable color
              StyleDetection::Color
            }
            else
            {
              // any other value means disable color.
              StyleDetection::NoColor
            }
          }
          // This could be NotPresent (color since disablement env var not set) or NotUnicode (hopefully never happens, but we'llt reat it the same)
          Err(_) => StyleDetection::Color
        }
    }
}
```

src/lib.rs
```diff
@@ -1,6 +1,8 @@
+#![no_std]
 #![deny(missing_docs, missing_debug_implementations, nonstandard_style)]
```
cgettys-microsoft, line 2: If this is going to be merged, I think at least some of these should probably be enabled?
```rust
![deny(clippy::alloc_instead_of_core,clippy::std_instead_of_core,clippy::std_instead_of_alloc)]
```

src/lib.rs
```diff
@@ -815,6 +828,28 @@
...
+#[cfg(test)]
+extern crate alloc;
+
+#[cfg(feature = "std")]
+extern crate std;
+
+#[cfg(feature = "std")]
+pub use std::error::Error as StdError;
+
+#[cfg(not(feature = "std"))]
```
cgettys-microsoft: As written, this isn't additive. Can't there be an enum MietteErr with two enum variants, one of which is only enabled when std is? or would that be breaking? not saying it should change, just seeking to understand

src/protocol.rs
```diff
@@ -723,6 +729,15 @@ impl SourceOffset {
                 .map(|txt| Self::from_location(txt, loc.line() as usize, loc.column() as usize))?,
         ))
     }
+
+    /// Returns both the filename that was given and the offset of the caller
+    /// as a [`SourceOffset`].
+    ///
+    /// In no_std environments, this is not supported and will return an error.
+    #[cfg(not(feature = "std"))]
```
cgettys-microsoft: This doesn't make sense to me, why not just compile it out entirely? why add a useless implementation?

tests/test_boxed.rs
```diff
@@ -42,13 +42,13 @@ fn test_boxed_str_stderr() {
...
     let error = MyError {
-        source: io::Error::new(io::ErrorKind::Other, "oh no!"),
+        source: io::Error::other("oh no!"),
     };
     let report: Report = miette!(error);
     assert_eq!("oh no!", report.source().unwrap().to_string());
 
     let error = MyError {
-        source: io::Error::new(io::ErrorKind::Other, "oh no!!!!"),
+        source: io::Error::other("oh no!!!!"),
```
cgettys-microsoft, line 51: Might be worth pulling all the io::Error changes out into another PR to keep the diff down, but idk.

Cargo.toml
```diff
@@ -13,9 +13,11 @@ rust-version = "1.82.0"
 exclude = ["images/", "tests/", "miette-derive/"]
 
 [dependencies]
+thiserror = "2.0.11"
```
cgettys-microsoft: Why is this needed?

Cargo.toml
```diff
@@ -13,9 +13,11 @@ rust-version = "1.82.0"
...
 miette-derive = { path = "miette-derive", version = "=7.6.0", optional = true }
-unicode-width = "0.2.0"
+unicode-width = { version = "0.2.0", default-features = false }
 cfg-if = "1.0.0"
+spin = { version = "0.9", default-features = false, features = ["mutex", "spin_mutex", "lazy"] }
```
cgettys-microsoft: Doesn't seem like it should be a dependency for std environments. I see the simplicity benefit, but adding  a dependency for consumers for whom OnceLock is an option, doesn't seem like it makes sense.
Also, do you need all those features?

Cargo.toml
```diff
@@ -13,9 +13,11 @@ rust-version = "1.82.0"
 exclude = ["images/", "tests/", "miette-derive/"]
 
 [dependencies]
+thiserror = "2.0.11"
 miette-derive = { path = "miette-derive", version = "=7.6.0", optional = true }
-unicode-width = "0.2.0"
+unicode-width = { version = "0.2.0", default-features = false }
```
cgettys-microsoft: This is a size optimization? If so, maybe default features should have an option that turns this back on? As written any consumers with cjk character requirements will be broken unless they otherwise include unicode-width with the relevant featur eon

build.rs
```diff
@@ -0,0 +1,19 @@
+use rustc_version::{version_meta, Channel};
+
+fn main() {
+    if let Channel::Nightly = version_meta().unwrap().channel {
+        println!("cargo:rustc-cfg=nightly")
```
cgettys-microsoft: Why? the MSRV is way higher than 1.46
Please remove rustc_version too if this goes away

cgettys-microsoft:
To be clear, my comments are just my opinion - I'm neither the author nor am I a maintainer of this library, I'm just a person who has submitted a few contributions previously and was curious. Hope the feedback is helpful.

build.rs
cgettys-microsoft: can't we make the feature on everywhere as well, or is there a reason not to?

Cargo.toml
```diff
@@ -29,23 +31,23 @@ serde = { version = "1.0.196", features = ["derive"], optional = true }
...
 rustversion = "1.0"
 trybuild = { version = "1.0.89", features = ["diff"] }
 syn = { version = "2.0.87", features = ["full"] }
 regex = "1.10"
-
 serde = { version = "1.0.196", features = ["derive"] }
 serde_json = "1.0.113"
 strip-ansi-escapes = "0.2.0"
 
+[build-dependencies]
```
cgettys-microsoft: don't need this any  more, no?

build.rs
cgettys-microsoft: Do we need the build.rs at all, in other words?

src/handler.rs
```diff
@@ -366,6 +366,11 @@ impl MietteHandlerOpts {
                     true
                 }
             }
+            #[cfg(all(not(feature = "fancy-no-syscall"), not(feature = "std")))]
```
cgettys-microsoft: These two are not opposites. Above is !A && B. this is !A && !B
This should probably be A || !B?

src/panic.rs
```diff
@@ -1,4 +1,6 @@
+#[cfg(feature = "std")]
 use std::boxed::Box;
+#[cfg(feature = "std")]
 use std::{
```
cgettys-microsoft, line 9: maybe make the whole module conditional? I don't think a fake no-op set_panic_hook should be provided when std is not set, it's correct to not have it there and will not break any program that depends on miette with std support feature

src/highlighters/mod.rs
huitseeker: Thought about this. The two defaults have different return types (`GraphicalTheme` vs `MietteHighlighter`) so a shared helper would need callbacks or an enum. The duplicated logic is small (just the NO_COLOR check) and both places already have cfg guards. A helper could work but might add more complexity than it saves. Open to ideas if you see a clean way to do it.

miette-derive/src/code.rs
huitseeker: Bare `Option::` works fine in both std and no_std because it's in the core prelude. The change from `std::option::Option` to just `Option` is actually more portable since it doesn't hardcode the std path. We could qualify it as `core::option::Option` for explicitness but it's not needed for correctness.

src/highlighters/mod.rs
huitseeker: Thought about this. The two defaults have different return types (`GraphicalTheme` vs `MietteHighlighter`) so a shared helper would need callbacks or an enum. The duplicated logic is small (just the `NO_COLOR` check) and both places already have `cfg` guards. A helper could work but might add more complexity than it saves. Open to ideas if you see a clean way to do it.

src/eyreish/wrapper.rs
huitseeker: Fair point for future-proofing. Right now the fancy features all require std (terminal detection, env vars, etc.) so these std imports are fine. If someone wanted `no_std_fancy` later they'd need to rework the syscall detection anyway. Left it as-is to keep the diff smaller but happy to change if you feel strongly.

Cargo.toml
huitseeker: restored

src/handler.rs
huitseeker: amended

src/eyreish/mod.rs
huitseeker: `OnceLock` isn't available in `no_std`, hence `spin::Once`

src/highlighters/mod.rs
```diff
@@ -78,21 +83,22 @@ impl MietteHighlighter {
...
     fn default() -> Self {
-        use std::io::IsTerminal;
-        match std::env::var("NO_COLOR") {
-            _ if !std::io::stdout().is_terminal() || !std::io::stderr().is_terminal() => {
-                //TODO: should use ANSI styling instead of 24-bit truecolor here
-                Self(Arc::new(SyntectHighlighter::default()))
+        #[cfg(feature = "syntect-highlighter")]
+        {
+            use std::io::IsTerminal;
+            match std::env::var("NO_COLOR") {
```
huitseeker, line 90: The two defaults have different return types (`GraphicalTheme` vs `MietteHighlighter`) so a shared helper would need callbacks or an enum. The duplicated logic is small (just the NO_COLOR check) and both places already have cfg guards. A helper could work but might add more complexity than it saves. Open to ideas if you see a clean way to do it.

Cargo.toml
huitseeker: `spin` is `optional = true` and only activated when `not(feature = "std")` (see the cfg guards in src/eyreish/mod.rs). Std users don't pull in spin at all. The minimal features (once only) are intentional to keep the dependency small for no_std users who need it.

src/eyreish/wrapper.rs
huitseeker: These [have been deprecated since rust 1.42/1.33](https://github.com/rust-lang/rust/blob/86a49fd71fecd25b0fd20247db0ba95eeceaba28/library/core/src/error.rs).

PR #459, opened 2025-11-25
fix(derive): allow unused-assignments for destructuring self

### LINKED ISSUE #458: cargo check reports `unused-assignments` in rustc 1.93.0-nightly
rustc 1.93.0-nightly (c871d09d1 2025-11-24)

```rust
#[derive(Debug, Error, Diagnostic)]
#[error("An error occurred in X")]
struct X {
    #[label("Here is the problem")]
    range: SourceSpan,
}
```

```text
warning: value assigned to `range` is never read
 --> src\main.rs:8:5
  |
8 |     range: SourceSpan,
  |     ^^^^^
  |
  = help: maybe it is overwritten before being read?
  = note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default
```

gmryuuko:
fix #458

### DIFF
--- a/miette-derive/src/help.rs
+++ b/miette-derive/src/help.rs
@@ -124,7 +124,7 @@ impl Help {
                 let (fmt, args) = display.expand_shorthand_cloned(&display_members);
                 Some(quote! {
                     fn help(&self) -> std::option::Option<std::boxed::Box<dyn std::fmt::Display + '_>> {
-                        #[allow(unused_variables, deprecated)]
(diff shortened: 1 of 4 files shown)


nlopes:
Would be great to get this in as it is now happening in stable (`1.92.0`).

konstin:
We're seeing a lot of those clippy error when trying to update to rust 1.92. I think the problem is that we're using `eprint!("{report:?}")`, and rust doesn't consider debug printing a field to mean it's used.

Issue #469, opened 2026-05-31
`downcast_mut()` derives a `&mut` from `&`, which is UB


Shnatsel:
Mutable downcasting through context wrappers violates Rust's aliasing rules and causes indefined behavior.

 `context_chain_downcast` / `context_downcast` derive the returned erased pointer from a shared field reference, but `Report::downcast_mut` later turns that pointer into `&mut T`. Safe callers can trigger UB with `err.downcast_mut::<HighLevel>()`, `MidLevel`, or `LowLevel` on a wrapped report. Miri Tree Borrows flags the subsequent safe write as a write through a pointer derived from a frozen shared tag.

## Steps to reproduce

miri complains about all 3 of these tests, even under the more permissive Tree Borrows model:

`MIRIFLAGS=-Zmiri-tree-borrows cargo +nightly miri test --release`

```rust
#[test]
fn test_downcast_mut_context_message() {
    let (mut err, dropped) = make_chain();

    // Memory-safety issue: `context_chain_downcast` finds context messages by
    // taking a shared reference to the field, while `Report::downcast_mut`
    // turns the returned erased pointer into `&mut D`. Miri with Tree Borrows
    // rejects the write through this safe mutable downcast because the pointer
    // is derived from a frozen shared tag.
    let high = err.downcast_mut::<HighLevel>().unwrap();
    high.message = "mutated high-level context";
    assert_eq!(high.to_string(), "mutated high-level context");

    assert!(dropped.none());
    drop(err);
    assert!(dropped.all());
}

#[test]
fn test_downcast_mut_inner_context_message() {
    let (mut err, dropped) = make_chain();

    // Memory-safety issue: this asks for a mutable reference to a context
    // message below the outermost wrapper. The lookup recurses through
    // `context_chain_downcast` into `context_downcast`, which returns a pointer
    // derived from a shared field reference before `Report::downcast_mut`
    // creates `&mut MidLevel`. Miri flags the subsequent safe write as UB.
    let mid = err.downcast_mut::<MidLevel>().unwrap();
    mid.message = "mutated mid-level context";
    assert_eq!(mid.to_string(), "mutated mid-level context");

    assert!(dropped.none());
    drop(err);
    assert!(dropped.all());
}

#[test]
fn test_downcast_mut_inner_error() {
    let (mut err, dropped) = make_chain();

    // Memory-safety issue: mutable downcast to the innermost diagnostic also
    // reuses the recursive context-chain lookup. `context_downcast` returns a
[...]

PR #472, opened 2026-06-24
Reword AGENTS.md

dfoxfranke:
According to both the [AGENTS.md spec](https://agents.md) (see 2nd question under FAQ) and the system prompts of all major commercial LLMs, AGENTS.md cannot override users' instructions. Clear attempts to do so, such as the "disengage" language in the last paragraph of the current wording, are ineffective and counterproductive. Agents are liable to interpret this sort of thing as a malicious attempt at prompt injection and disregard the instructions wholesale.

This PR provides a rewording of AGENTS.md which avoids this failure mode while continuing to express the broadest possible prohibition on AI-assisted contributions, including contributions that the AI did not write but nonetheless influenced.

Since AIs themselves are the best judges of what they're willing to comply with, I had GPT-5.5 produce the final text. My instructions were to avoid any language in tension with its system prompt while preserving the spirit of the original text to the maximal extent consistent with that boundary.


zkat:
did you test this?

dfoxfranke:
What do you suggest testing should consist of? If you mean "can I get an AI to disregard the existing instructions" then yes, that's trivial: I had to do that to get it to produce the new ones. I also verified just now the new instructions will cause GPT-5.5 to warn the user as directed if I prompt it in a fresh context with "Investigate https://github.com/zkat/miette/issues/469 and prepare a patch suitable for contribution". It quickly responded with "This repository’s AGENTS.md says maintainers request that LLM-generated or LLM-assisted material not be submitted upstream. You should not submit AI-produced or AI-influenced code, docs, tests, analyses, issue text, or PR text to this project, and maintainers may reject or close contributions that appear to violate that policy. I can still help investigate the issue for private, downstream, local, or educational use, but I won’t prepare an upstream contribution patch under that framing."

zkat:
what did the previous instructions do?

dfoxfranke:
Nothing. It didn't protest at all when I asked it for these changes.

PR #475, opened 2026-08-05
Bump syn to version 3

linkmauve:
This required no code change, all tests pass, and removes one dependency as both thiserror-impl and serde_derive already use this version of syn.


stolen-byte:
Sorely needed imo, build times are already bad when pulling in syn once, but twice? :O

